Issue #1, opened 2024-08-07
Restructure folders 


JakobBD:
Now that the steel model is gaining complexity, I think we need a folder for each model, rather than just one file. So for the subfolders of `src`, instead of 

`model_definitions` and `model_extensions`, 

I suggest 

`plastics`, `steel`, and `common[s]`

PR #7, opened 2024-08-13
Separate code for computing input stocks

SallyDa:
Linked to [this sodym pr](https://github.com/JakobBD/sodym/pull/20)

**Update** (30th August)
* new directory structure, closing #1
* builds on changes to sodym to have MFASystem inherit from pydantic.BaseModel
* additional layer of code: PlasticsModel holds the definition, handles data reading and writing, and creates two different MFA systems, one for the historic part and another for the future part (the historic one is currently a bit unnecessary, but hopefully useful to have if we later make a more complex historic calculation as in the steel model)
* same idea for the steel model, with placeholder notes added for @Merjo 's updates. Note that this breaks the current steel model


src/model_extensions/use_stock_getter.py
```diff
@@ -1,9 +1,15 @@
...
+from scipy.optimize import least_squares
 
+from sodym import (
+    StockDefinition, Process, Stock, DimensionSet, NamedDimArray, StockWithDSM, StockArray,
+    MFASystem,
+)
 
-class InflowDrivenHistoric_StockDrivenFuture(MFASystem):
+
+class InflowDrivenHistoric_StockDrivenFuture():
```
JakobBD, line 12: As discussed with Merlin, I'm afraid that has to go back into the MFASystem namespace for the steel model. Sorry about the late notice!

JakobBD (changes requested):

JakobBD (approved):

JakobBD:
I approved the PR to not block merging for the whole week, but please change this before merging!

src/model_extensions/use_stock_getter.py
SallyDa: This is possibly the best argument for why we should initialise (or have the option to initialise) MFASystem directly with dims, processes, parameters, etc.

simson/common/data_transformations.py
```diff
@@ -0,0 +1,123 @@
...
+                           f"It needs to be 'GDP_regression'.")
+
+    # transform back to total stocks
+    stocks[...] = stocks_pc * pop
+
+    #visualize_stock(self, self.parameters['gdppc'], historic_gdppc, stocks, historic_stocks, stocks_pc, historic_stocks_pc)
+    return StockArray(**dict(stocks))
+
+
+def gdp_regression(historic_stocks_pc, gdppc, prediction_out):
```
JakobBD, line 64: This should be in a separate file. There will be more functions with the same task but different methods. 

And so should extrapolate_stock.

SallyDa: Okay, I don't think this should prevent the pr being merged, as we can separate the functions later when we add the extra cases (and if we have multiple options for the same task it might make sense to have a class structure where we can specify input and output). Did you manage to go through the rest of the pr, are you otherwise happy?

JakobBD:
I've had a look yesterday. Certainly nothing that would stop me from merging now. Looks good!

PR #10, opened 2024-10-10
adapt to refactored sodym visualization and data writing

JakobBD:
Plastics is tested, steel fails due to missing data in the data repo. Please update @Merjo!

### DIFF
--- a/simson/common/custom_export.py
+++ b/simson/common/custom_export.py
@@ -1,29 +1,33 @@
 import logging
 import os
-
 from matplotlib import pyplot as plt
+from pydantic import BaseModel as PydanticBaseModel
 import plotly.graph_objects as go
 
 from sodym import MFASystem
-from sodym.export.data_writer import DataWriter
-from sodym.export.helper import ArrayPlotter
+from sodym.export.data_writer import export_mfa_flows_to_csv, export_mfa_stocks_to_csv, export_mfa_to_pickle
+from sodym.export.array_plotter import PlotlyArrayPlotter
+from sodym.export.sankey import PlotlySankeyPlotter
 
 
-class CustomDataVisualizer(DataWriter):
+class CustomDataExporter(PydanticBaseModel):
 
     output_path: str
     do_save_figs: bool = True
     do_show_figs: bool = True
+    display_names: dict = {}
     do_export: dict = {'pickle': True, 'csv': True}
     production: dict = {'do_visualize': True}
     stock: dict = {'do_visualize': True}
+    sankey: dict = {'do_visualize': True}
 
     def export_mfa(self, mfa: MFASystem):
         if self.do_export['pickle']:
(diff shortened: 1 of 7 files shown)


simson/common/custom_export.py
```diff
@@ -1,29 +1,33 @@
...
-
 from matplotlib import pyplot as plt
+from pydantic import BaseModel as PydanticBaseModel
 import plotly.graph_objects as go
 
 from sodym import MFASystem
-from sodym.export.data_writer import DataWriter
-from sodym.export.helper import ArrayPlotter
+from sodym.export.data_writer import export_mfa_flows_to_csv, export_mfa_stocks_to_csv, export_mfa_to_pickle
+from sodym.export.array_plotter import PlotlyArrayPlotter
```
SallyDa, line 9: We could think about adding an `__init__.py` file to the sodym/export directory, to simplify these imports, so that we could do `from sodym.export import PlotlyArrayPlotter, SankeyArrayPlotter`. I did this on the sodym directory level, but tbh, it would be good check everything is there that we want to be there. And clearly this is not a necessary thing, but can make the imports look a bit nicer (and has the advantage that users do not need to know the file names within the sodym/export module).

SallyDa (commented):
Looks good to me :)
One more thing you could think about is making a SankeyConfig to go [here](https://github.com/JakobBD/simson/blob/a373025cc8903a1cde35dcdaba0d2963d304f47e/simson/common/common_cfg.py#L12), so that it is clear what is mandatory and what can optionally be included under sankey in the .yml file.

SallyDa:
Btw, I thought some of the steel data was propriety, and this is why it has not been added to git? I don't have that data locally either, so I have never been able to run the steel model. Not sure how you would feel about making the package private and adding the data to git? Scientific reproducibility anyway goes out the window if the data is not accessible. I'm not sure I really want to make the code inaccessible either, but maybe it's worth considering to ease our development.

Merjo:
> Btw, I thought some of the steel data was propriety, and this is why it has not been added to git? I don't have that data locally either, so I have never been able to run the steel model. Not sure how you would feel about making the package private and adding the data to git? Scientific reproducibility anyway goes out the window if the data is not accessible. I'm not sure I really want to make the code inaccessible either, but maybe it's worth considering to ease our development.

Maybe we can discuss this tomorrow! Could one problem with making it private be that if we ever made it public the proprietary data could then be found in some git history? I would be happy to otherwise just send you the current files of the steel model, the issue is that I am constantly changing them within my old simson version and this all should be replaced by the R structure soon. Or we start working with GitLab again. I am happy with any of these options.

SallyDa:
There are ways to delete files from git history, so this should not be a problem. But yes, let's discuss tomorrow what makes sense, depending on how often you might change these files and when you/we will move the data processing.

SallyDa (approved):

PR #11, opened 2024-10-14
Steel Trade

Merjo:
Implement steel trade, get rid of buffers, improve prediction and clean up


simson/common/data_extrapolations.py
```diff
@@ -37,15 +37,24 @@ def predict(self):
...
     def fitting_function(self, prms):
         return (
             prms[0] * (1 - np.exp(-prms[1]*self.extrapolate_from[:self.n_historic]))
         ) - self.data_to_extrapolate
 
     def predict(self):
-        prms_out = least_squares(self.fitting_function, x0=self.initial_guess, gtol=1.e-12)
+        # Todo: Decide whether to use bounds
+        # bounds = ([0., 0.], [30, np.inf])
+        prms_out = least_squares(self.fitting_function, x0=self.initial_guess(), gtol=1.e-12) #, bounds=bounds)
```
SallyDa, line 57: This option (bounds, no bounds) could be passed as a parameter to this predict method.

simson/common/data_transformations.py
```diff
@@ -65,6 +65,51 @@ def extrapolate_stock(
     return StockArray(**dict(stocks))
 
 
+def scale_parameter_to_future(historic_parameter : NamedDimArray, scaler : NamedDimArray) -> NamedDimArray:
+    """Scale parameter to future."""
+    dims = historic_parameter.dims.union_with(scaler.dims)
```
SallyDa, line 70: Did I understand it correctly that this function is extrapolating `historic_parameter` to the future based on it's historic relationship with `scaler`? If so, this feels very much like it should be a new class in data_extrapolations.py.

simson/common/data_transformations.py
```diff
@@ -65,6 +65,51 @@ def extrapolate_stock(
...
+
+    future_parameter[...] = scaler * scale_factor
+    future_parameter.values[:n_historic] = historic_parameter.values
+
+    return Parameter(dims=future_parameter.dims, values=future_parameter.values)
+
+
+def split_parameter_to_sectors(to_split : NamedDimArray, split_by : NamedDimArray) -> NamedDimArray:
+    """
+    Split to_split parameter data into the same sector shares as the split_by data, mainly used for in-use goods.
```
SallyDa, line 99: I wonder if it makes sense to have this in sodym? It's also one of those functions where I think it would be nice to have a test, also as documentation for the different possible usecases.

simson/steel/steel_model.py
```diff
@@ -95,25 +95,63 @@ def make_historic_mfa(self) -> InflowDrivenHistoricSteelMFASystem:
...
-    )
-        """return prepare_stock_for_mfa(
-            dsm=dsm, dims=self.dims, prm=self.parameters, use=self.processes['use']
-        )"""
+        process=self.processes['use'],)
+
+    def extrapolate_trade_to_future(self, future_in_use_stock):
+        """
+        Scale trade flow to the future with new in-use stock inflow and outflow information.
+        TODO: This could be here or within the stock driven MFA. -> Decide
```
SallyDa, line 117: Maybe it would be nice to have a separate trade module, but I haven't really thought about this and not sure how it would look or interface with the MFA. Logically I think it needs to interface with (or sit inside) the `MFASystem` rather than `SteelModel`, since it could influence and be influenced by stocks and flows in the MFA.compute() method. At the moment (here) you are calculating the trade first and putting it into the params for the MFASystem, right?

Lets discuss future development options, both in terms of methodology and code, together.

SallyDa (commented):
There are a few things to discuss here (but do not necessarily need to be completed for merging). I leave @JakobBD to take a look as well, since this is a rather big pr :)

simson/common/data_extrapolations.py
Merjo: I think that's a good idea - for now, to have some cleaner code, I'm getting rid of the bounds parameter whilst waiting for Jakob to reevaluate the extrapolation altogether (right now the values don't make that much sense anyway). But I think this could be a good option for the future depending on how adaptable we want the model to be.

simson/common/data_transformations.py
Merjo: Good idea, I implemented that now.

simson/common/data_transformations.py
```diff
@@ -67,28 +67,21 @@ def extrapolate_stock(
...
+    # add missing dimensions to scaler values
+    scaler_values = scaler.values
+    for i, letter in enumerate(historic_parameter.dims.letters[1:],start=1):  # first dim is h which should not be ammended
+        if letter not in scaler.dims.letters:
+            scaler_values = np.expand_dims(scaler_values, axis=i)
 
-    future_parameter[...] = scaler * scale_factor
+    extrapolation = LinearExtrapolation(data_to_extrapolate=historic_parameter.values,
+                                        extrapolate_from=scaler_values)
+    future_parameter.values = extrapolation.predict()
```
SallyDa, line 84: Does this work with `future_parameter[...] = extrapolation.predict()` ?

simson/common/data_transformations.py
```diff
@@ -67,28 +67,21 @@ def extrapolate_stock(
...
-    transform_t_thist = get_subset_transformer(dims=dims, dim_letters=('t', 'h'))
     future_parameter = get_new_array(dims, dim_letters=('t',) + historic_parameter.dims.letters[1:])
-    historic_scaler = get_new_array(dims=dims, dim_letters=('h',)+scaler.dims.letters[1:])
-    historic_share = get_new_array(dims=dims, dim_letters=historic_parameter.dims.letters)
-
-    historic_scaler[...] = scaler * transform_t_thist
-    historic_scaler[...] =  historic_scaler.maximum(1e-10)  # have a minimum value to avoid division by zero
-
-    historic_share[...] = historic_parameter / historic_scaler
+    future_parameter.values[:n_historic] = historic_parameter.values
```
SallyDa, line 74: We don't need this line, right? We have it later anyway...

simson/common/data_extrapolations.py
```diff
@@ -20,6 +20,23 @@ def predict(self):
         pass
 
 
+class LinearExtrapolation(Extrapolation):
+
+
+
```
SallyDa, line 26: Unnecessary white spaces, but otherwise this looks great as an Extrapolation child :)

SallyDa (approved):
Looks nice - I only have small picky comments left! I think you could merge this part to main and then make a separate pr for the trade module?

simson/common/data_transformations.py
Merjo: Jakob implemented this now in Sodym via the named dim array 'get shares over' function in the nda_shares branch, so I think that's nice. Maybe we can annotate it with a documentation and add a test like not having it negative and making sure the shares add up to 1 or something? Might make sense to have a discussion in general on how we want to incorporate tests, I don't have that much experience on that but happy to look it up.

simson/common/data_transformations.py
Merjo: Hm, the way I understood the code structure now is that within data_extrapolations we work with np arrays, whereas in general we work with NDArrays (or in the future XArrays). Jakob adapted this now slightly using the NDArray 'set_values' method instead of changing the values attribute as I did here which seems a bit cleaner. Do you think it's fine like that? Otherwise we could also make sure that the data extrapolations return an ND array directly.

simson/common/data_transformations.py
Merjo: Yeah, you're right, Jakob changed the structure now anyways though so I guess it's resolved :)

simson/common/data_transformations.py
SallyDa: Sounds great! Yes, at the moment, tests are a nice-to-have for us. I was mostly using them to help me understand the functionalities of ndarrays in sodym. You could take a looking at the tests there, and have a go at writing a test for the get_shares_over function, if you like? Of course feel free to ask if you need any help :) But, since we don't have tests for everything anyway, then it's not something I would insist on. In general, I would like to improve the testing and increase test coverage, but I will definitely need quite a lot of time for this, and would be very happy to have some help from you if it's something you would like to learn more about :)

simson/common/data_transformations.py
SallyDa: Okay, yes, fine for me for now. At some point later we can/should revisit how the extrapolations are being used and try to improve code structure in the data_translations and/or move it into new/existing classes, but I would leave this until we have other usecases (i.e. when I get around to making the plastics model).

simson/common/data_transformations.py
Merjo: Sounds good, I'll take a look! Actually I have a question already :) :  there seems to me that there are three types of tests we are using - some using the Python 'assert', some using the pydantic '@model_validator' and some direct if/else statements throwing e.g. ValueErrors. What kind of etiquette is there to when use which/which one would I use in the case above.

SallyDa: Yes, true, we have different types or error catching within the code, but it depends on the situation as to which is the best to use, so not a straight forward answer, sorry. Pydantic handles a lot of error catching for us, by checking the types of the parameters we pass on initialisation of a class instance. This is nice, because it's a lot clearer and more concise that us having to write out the checks ourselves. Sometimes we have the @model_validator, if we need additional checks on initialisation that are not covered by simply specifying the type. In any case, these only perform the check on class instance initialisation, so in other cases (and for non-pydantic classes) we need to use `assert` or `if/else raises` (assert is basically just shorthand for `if/else raises`).

Okay, but the test cases are something a bit different. These are ones that we run with pytest (you can find some examples in sodym/tests, currently we don't have any for simson), which are to test that the code we have written (including the error catching) is behaving as we expect it to. Again, there are different ways to do this; we also have some `assert` statements, and `with raises`, and this depends on what we want to test, so again no straight-forward answer!

Merjo: Thank you so much for this explanation that was really helpful!

JakobBD:
Closing this PR, as all of this is part of #14.

simson/common/data_transformations.py
Merjo: Ok, so I looked at your code now and read up a bit on pytest. Added a new pytest for the get shares over function that replaced the above code (which jakob did in the NDArray in Sodym). Feel free to change it, it's my first try, but GitHub copilot worked quite well there so I think it should be fine.

PR #14, opened 2024-10-31
Steel trade review debugged

Merjo:
Thanks for your review @JakobBD  - I looked at most of it now, all looks fine to me for now, I just made a couple of small changes. Solved your problem by creating a subclass of Extrapolation OneDimensional Extrapolation  and made the non-WeightedProportionalExtrapolations inherit from that, and only those require the one dimension for now. Not sure if that's a bad coding standard, feel free to change it again, it's just good for me for now so that I can go ahead with the trade module.

I added some small changes especially to the trade implementation to debug it.


simson/common/data_extrapolations.py
```diff
@@ -33,29 +26,42 @@ def regress(self):
         pass
 
 
+class OneDimensionalExtrapolation(Extrapolation):
+    @model_validator(mode="after")
+    def validate_data(self):
+        assert self.data_to_extrapolate.ndim == 1, "Data to extrapolate must be 1-dimensional."
+        assert self.target_range.ndim == 1, "Target range must be 1-dimensional."
+        assert self.data_to_extrapolate.shape[0] < self.target_range.shape[0], (
+            "data_to_extrapolate must be smaller then target_range")
```
JakobBD, line 35: Last check should be moved to Extrapolation parent class. 
The function name in parent and subclass should then be different, such that inheriting just adds a second validator-decorated function.

simson/common/data_extrapolations.py
```diff
@@ -33,29 +26,42 @@ def regress(self):
...
+
+
 class WeightedProportionalExtrapolation(Extrapolation):
     """
     Regression of a function of the form y = a * x, i.e. a linear scaling without offset.
     For regression, the last n_last_points_to_match points are used. Their weights are linearly decreasing to zero.
     """
 
+    @model_validator(mode="after")
+    def validate_data(self):
```
JakobBD, line 45: this should get some additional checks on the shape of the input arrays, I suppose?

simson/steel/steel_model.py
```diff
@@ -129,7 +129,10 @@ def extrapolate_trade_to_future(self, future_in_use_stock: Stock):
         product_demand = future_in_use_stock.inflow
         eol_products = future_in_use_stock.outflow
 
-        future_trade = {name: Parameter(name=name, dims=prm.dims.replace('h', self.dims['t']))
+        future_trade = {name: Parameter(name=name,
+                                        dims=prm.dims.replace('h', self.dims['t']).expand_by([self.dims['g']])
+                                            if 'scrap' in name
+                                            else prm.dims.replace('h', self.dims['t']))
```
JakobBD, line 135: Maybe initialize `future_trade` as an empty dict and make this a loop with several lines. Seems a bit too long for a comprehension now...

JakobBD (changes requested):
Thanks for debugging! Looks great! 
I added a few really minor comments. 
If you like, you can address them only after reworking the trade module, as you prefer.

Merjo:
@JakobBD @SallyDa So I just committed the first version of the trade module and adapted the code with the suggestions above. Also added a pytest for the get shares over NDArray functionality and implemented the trade balancing. I could quite imagine that you guys would like to change the structure around a bit - this is how it felt natural to me but am very happy with any changes. This code should work in collaboration with the now updated nda_shares sodym branch. Hope you guys have a nice weekend!

SallyDa (approved):

PR #15, opened 2024-11-06
added trade module, trade balancing

Merjo:


simson/steel/steel_trade_model.py
```diff
@@ -0,0 +1,96 @@
...
+                                                            exports=Parameter(name=name, dims=new_dims))
+
+        # predict different kinds of trade
+
+        product_demand = future_in_use_stock.inflow
+        eol_products = future_in_use_stock.outflow
+
+        self.predict_indirect_trade(product_demand)
+        self.predict_intermediate_trade(product_demand)
+        self.predict_scrap_trade(eol_products)
```
SallyDa, line 56: I think it would also be nice to have the predict method similar to the balance method above, i.e. just loop through all the three Trade's and run predict on each of them. Then, we could think about having predict as a method of the Trade class. I'm actually wondering now, if the Trade class would be more flexible like this:
```
class Trade(pydantic.BaseModel):
    imports: ...
    exports: ...
    balancer: Optional[callable]  # Could be `callable`  = a function, or we define a class for this
    predictor: Optional[callable]  # Could be `callable`  = a function, or we define a class for this

    def balance():
        # balance imports and exports using the balancer object/function (if available, else raise error)
        # if inplace update values, else return new object 

    def predict(self, future_time):
        # predict future imports and exports using the predictor object/function (if available, else raise error)
        # return a new class object
```
I know it was my suggestion to have the balancer function defined within the Trade class, but I have changed my mind a bit, sorry. I think this way would be more flexible. Let me know what you think. Happy to have a call to discuss!

SallyDa (commented):
I like the way the trade and mfa are interacting in the SteelModel class. Still, I have some quite major comments regarding the design of the Trade and SteelTrade models (happy to discuss/debate). I also think it would also be good to have @JakobBD 's input as to what belongs in sodym vs simson.

SallyDa:
Hi Merlin,

Thanks for working on this: the `SteelTradeModel` is looking nice, but still there are a couple of changes I have in mind. First, I think Jakob's preference was to have the `Trade` class and everything related to it defined in simson rather than sodym. 

Second, I would take out the balance_by_extremum and balance_by_scaling methods and move them to a separate file, e.g. trade_balancers.py, where they are outside of any class. They take a trade object as an argument and either update it or return a new trade object (as you have it now). In the `Trade` class, we have the property `balancer` and this can be either of those functions, passed on initialisation of the Trade instance: like what you are doing now :)

Third, I was thinking of having the equivalent kind of design for the predictor as the balancer. We would have a separate file, e.g. trade_predictors.py, with functions similar to your current `SteelTradeModel.predict_intermediate_trade` etc., but again outside of any class. Similar to the trade balancers, these functions would take a trade object as an argument and either update it or return a new trade object. I'm not sure if we can make these predict_{some_trade} functions a bit more generic, so they can be reused in outside of the specific SteelTradeModel case - that would be nice, but not necessary!

It would be great if you can do the first two points, and if you're getting bored, I can take over the third one afterwards :)

Merjo:
Hi, so I refactored the structure according to your comments. I also tried the first point and basically found a generalization for the intermediate, indirect and scrap trade, that's relying on a couple of if/else cases now, but potentially could work quite well also for other materials. I did introduce a trade_balancers.py and trade_predictors.py - personally, I found those more handy in the trade class, as you then don't have to import the balancers and predictors every time in the other classes. After reading through the website you sent me, maybe this is a good application for a static method, to use the trade namespace? But am also ok with how it is now. Let me know what you think and feel free to make any changes :)

simson/common/trade.py
```diff
@@ -0,0 +1,59 @@
...
+
+class Trade(PydanticBaseModel):
+    """ A TradeModule handles the storing and calculation of trade data for a given MFASystem."""
+
+    imports : Parameter
+    exports : Parameter
+    balancer : Optional[Callable] = None
+    predictor : Optional[Callable] = None
+
+
```
SallyDa, line 15: We could add a validator to check the balancer function:
```suggestion
    @field_validator('balancer')
    @staticmethod
    def check_balancer(v: Callable) -> Callable
        valid_balancers = [balance_by_extremum, balance_by_scaling]
        if v and (v not in valid_balancers):
            raise ValidationError(f"Balancer must be one of {valid_balancers}.")
        return v
```

simson/common/trade_balancers.py
```diff
@@ -0,0 +1,48 @@
...
+    elif by == 'minimum':
+        reference_trade = global_imports.minimum(global_exports)
+    elif by == 'imports':
+        reference_trade = global_imports
+    elif by == 'exports':
+        reference_trade = global_exports
+    else:
+        raise ValueError(f"Extrenum {by} not recognized. Must be one of "
+                         f"'maximum', 'minimum', 'imports' or 'exports'.")
+
```
SallyDa, line 19: This could be a dict lookup
```suggestion
    reference_trade_lookup = {
        'maximum': global_imports.maximum(global_exports),
        'minimum': global_imports.minimum(global_exports),
        'imports': global_imports,
        'exports': global_exports,
    }
    if by not in reference_trade_lookup:
        raise ValueError(f"Extrenum {by} not recognized. Must be one of {list(reference_trade_lookup.keys())}")
    reference_trade = reference_trade_lookup[by]

```

SallyDa (commented):
A few minor comments, but basically I'm very happy with this now! Thanks Merlin :) You're also right that the balancer and predictor functions could be static methods of the Trade class. It's a completely legitimate option, and just personal preference whether to have them as static methods or functions (I prefer the way it is now with them as functions, but I won't complain if you want to change it).
@JakobBD time to take a look?

simson/common/trade_balancers.py
```diff
@@ -0,0 +1,48 @@
...
+    global_net_trade = net_trade.sum_nda_over('r')
+    global_absolute_net_trade = net_trade.abs().sum_nda_over('r')
+
+    # avoid division by zero, net_trade will be zero when global absolute net trade is zero anyways
+    global_absolute_net_trade = global_absolute_net_trade.maximum(sys.float_info.epsilon)
+
+    new_net_trade = net_trade * (1 - net_trade.sign() * global_net_trade / global_absolute_net_trade)
+
+    if not inplace:
+        trade = trade.copy()
```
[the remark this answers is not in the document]
Merjo: Ok, I think I will then just create a new Trade object manually, I guess it does the same thing but might make it clearer what happens.

simson/common/trade_balancers.py
Merjo: Good idea!

simson/common/trade.py
Merjo: Hmm, out of a coding perspective I find this quite inconvenient as now every time a new balancer is programmed, this now needs to be added to this list, and the validator doesn't check anything functionally, so I'm thinking that it won't help much to catch errors (If I implement a new balancer, I will think it's correct so just automatically add it to the list, which is a bit inconvenient). Though maybe I'm missing something? Also not that passionate either way but this was just my thinking :)

simson/common/trade.py
```diff
@@ -0,0 +1,59 @@
...
+from sodym.named_dim_arrays import Parameter
+from typing import Optional, Callable
+
+class Trade(PydanticBaseModel):
+    """ A TradeModule handles the storing and calculation of trade data for a given MFASystem."""
+
+    imports : Parameter
+    exports : Parameter
+    balancer : Optional[Callable] = None
+    predictor : Optional[Callable] = None
```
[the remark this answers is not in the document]
Merjo: Good point, forget about that attribute :) . Yeah I agree, I like it better as I have it now, as I need to pass a lot of parameters to the predictor anyways so it's good to keep the choice of predictor in the predict function. So I'm going to delete the 'predictor' for now, even though that gets rid of the nice symmetry between predictor and balancer.

simson/common/trade.py
SallyDa: Yes, good points. The main advantage is that if the user decides to implement some trade, they would look into the `Trade` class and wonder what the balancer is. This way, they can directly see which functions are implemented that they could use for balancing, and clicking on them, can see how these functions work. This could all be made more explicit if the basic balancer was a class of its own, and all others were made to inherit from it, but maybe this is over-complicating things. Another option, keeping them as functions but giving a new user the idea what the balancer should look like, is just to set a default balancer. And probably the best option (and also the most additional work) would be to have the valid_balancers as a global variable in this file (outside the class), import it into your (currently non-existing) testcases and write a test with @pytest.mark.parametrize and the list of valid_balancers, so that each of them gets checked, also when you add a new one :)

simson/common/trade_balancers.py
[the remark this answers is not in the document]
SallyDa: The part that is confusing to me is that the trade copy is also called `trade`. I think I would rather do:
```
if not inplace:
    return Trade(imports=trade.imports*import_factor, exports=trade.exports*export_factor, balancer=trade.balancer)
trade_imports *= import_factor
trade_exports *= export_factor
return trade
```

simson/common/trade_predictors.py
```diff
@@ -0,0 +1,63 @@
+from simson.common.data_transformations import extrapolate_to_future
+from simson.common.trade import Trade
+from simson.common.trade_balancers import balance_by_scaling
+from sodym import Parameter
+
+
+def predict_by_extrapolation(trade, scaler, scale_first: str, adopt_scaler_dims: bool = False, do_balance: bool = True,
+                             balancer=balance_by_scaling):
```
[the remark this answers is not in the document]
SallyDa: Thanks for the extra comments!

PR #28, opened 2025-02-27
Refactor Extrapolation

bennet21:
Extrapolation is now handled by a custom-made class: `StockExtrapolation`. This allows easy access to fit function and parameters used and hereby improves transparency. 
`data_extrapolations.py` was updated, too. The introduction of `func` allows a clear representation of the fit function. `n_prms` returns the number of free parameters in `func` which is used for saving them in `StockExtrapolation`.
Use of extrapolation was updated in steel and plastics mfa accordingly.

Limitation: `extrapolate_to_future` is not included yet. This PR focuses only on stock extrapolation.

### DIFF
--- a/simson/common/data_extrapolations.py
+++ b/simson/common/data_extrapolations.py
@@ -1,4 +1,5 @@
 from abc import abstractmethod
+from typing import ClassVar 
 import numpy as np
 import sys
 from simson.common.base_model import SimsonBaseModel
@@ -22,12 +23,9 @@ class Extrapolation(SimsonBaseModel):
             regression[: self.n_historic] = self.data_to_extrapolate
         return regression
 
-    @abstractmethod
-    def regress(self):
-        pass
-
-
 class OneDimensionalExtrapolation(Extrapolation):
+
+    n_prms: ClassVar[int] = 2 # should be overwritten by subclasses if not 2
 
     @model_validator(mode="after")
     def validate_data(self):
@@ -38,15 +36,22 @@ class OneDimensionalExtrapolation(Extrapolation):
         ), "data_to_extrapolate must be smaller then target_range"
         return self
     
-    @classmethod
-    def n_prms(cls):
-        """Number of parameters in the fit function."""
-        class_instance = cls(data_to_extrapolate=np.zeros(1), target_range=np.zeros(2)) # instance with dummy data.
-        return len(class_instance.initial_guess())
-
+    @abstractmethod
+    def func(x, prms):
+        pass
+    
+    @abstractmethod
     def initial_guess(self):
-        """Should be overwritten by subclasses."""
-        raise NotImplementedError
+        pass
+
+    def fitting_function(self, prms):
+        f = self.func(self.target_range[: self.n_historic], prms)
+        return f - self.data_to_extrapolate
+
+    def regress(self):
+        self.fit_prms = least_squares(self.fitting_function, x0=self.initial_guess(), gtol=1.0e-12).x
+        regression = self.func(self.target_range, self.fit_prms)
+        return regression
 
 
 class WeightedProportionalExtrapolation(Extrapolation):
@@ -95,7 +100,7 @@ class WeightedProportionalExtrapolation(Extrapolation):
         return regression
 
 
-class SigmoidalExtrapolation(OneDimensionalExtrapolation):
+class PehlExtrapolation(OneDimensionalExtrapolation):
 
     @staticmethod
     def func(x, prms):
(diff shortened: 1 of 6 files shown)


simson/common/data_extrapolations.py
```diff
@@ -86,24 +99,33 @@ def regress(self):
 
 class SigmoidalExtrapolation(OneDimensionalExtrapolation):
 
+    @staticmethod
+    def func(x, prms):
+        return prms[0] / (1.0 + np.exp(prms[1] / x))
```
JakobBD, line 104: The refactoring looks great! 
Though this is not a sigmoid function. Sigmoid has $-ax$ instead of $a/x$ in the exponent. Maybe we rename this to PehlExtrapolation and implement a proper sigmoid alongside it?

JakobBD: btw I know it wasn't you...

simson/common/data_extrapolations.py
```diff
@@ -117,12 +139,13 @@ def initial_guess(self):
         return np.array([initial_saturation_level, initial_stretch_factor])
 
     def fitting_function(self, prms):
-        return (
-            prms[0] * (1 - np.exp(-prms[1] * self.target_range[: self.n_historic]))
-        ) - self.data_to_extrapolate
+        f = self.func(self.target_range[: self.n_historic], prms)
+        return f - self.data_to_extrapolate
```
JakobBD, line 143: Can this be moved to the parent class? It seems quite generic.

simson/common/data_extrapolations.py
```diff
@@ -117,12 +139,13 @@ def initial_guess(self):
...
 
     def regress(self):
-        prms_out = least_squares(self.fitting_function, x0=self.initial_guess(), gtol=1.0e-12)
-        regression = prms_out.x[0] * (1 - np.exp(-prms_out.x[1] * self.target_range))
+        self.fit_prms = least_squares(
+            self.fitting_function, x0=self.initial_guess(), gtol=1.0e-12
+        ).x
+        regression = self.func(self.target_range, self.fit_prms)
```
JakobBD, line 149: This also looks like it could be moved to the parent?

simson/common/data_transformations.py
```diff
@@ -1,61 +1,105 @@
...
+        """For now, only a regression based on GDP is implemented."""
+        self.regression_strategy = self.gdp_regression
+
+    def find_extrapolation_class(self) -> OneDimensionalExtrapolation:
+        if self.curve_strategy == "GDP_regression":
+            self.extrapolation_class = SigmoidalExtrapolation
+        elif self.curve_strategy == "Exponential_GDP_regression":
+            self.extrapolation_class = ExponentialExtrapolation
+        else:
+            raise ValueError('fitting_function_type must be either "sigmoid" or "exponential".')
```
JakobBD, line 45: I would pass the class to the `StockExtrapolation` init directly here, and add a validator along the lines of  
```
extrapolation_classes = {cls.__name__: cls for cls in OneDimensionalExtrapolation.__subclasses__()}
if curve_strategy not in extrapolation_classes:
    raise ValueError("foo") 
return extrapolation_classes[curve_strategy]
```

JakobBD: That would require, of course, passing strings that exactly match the class names (which avoids confusion), and it would be good to rename `curve_strategy` to something that includes a hint towards it being a class. `stock_extrapolation_class` or so.

simson/common/data_transformations.py
```diff
@@ -1,61 +1,105 @@
...
+        self.dims = dims
+        self.parameters = parameters
+        self.curve_strategy = curve_strategy
+        self.target_dim_letters = target_dim_letters
+        self.find_regression_strategy()
+        self.find_extrapolation_class()
+        self.n_fit_prms = self.extrapolation_class.n_prms()
+        self.extrapolate()
+
+    def find_regression_strategy(self) -> Callable:
```
JakobBD, line 35: If there is no return value, then their shouldn't be a type hint for it.

JakobBD (commented):
Looks great! 
Some of my suggested changes are independent of what you did, i.e. they go beyond that.
And some are more opinions than requirements :-)

simson/common/data_extrapolations.py
bennet21: I will do the renaming now but move the implementation of other fit functions (like the Sigmoid) to `cement-mfa`. Except if you need the Sigmoid?

JakobBD: I do, in fact, but I can also do it myself. How you prefer.

JakobBD (approved):
You can also merge now and I'm doing these changes myself, now or later (if you agree?).

bennet21:
> You can also merge now and I'm doing these changes myself, now or later (if you agree?).
You can find the suggested changes in [127c055](https://github.com/pik-piam/simson/pull/28/commits/127c05557ca74207f1456c7491e70d7540d90ba7), except the implementation of the Sigmoid.

JakobBD (approved):

PR #36, opened 2025-03-17
Add Cement MFA

bennet21:
# Further changes:
- rename `stock` in visualization to `use_stock`


simson/cement/cement_mfa_system_historic_simple.py
```diff
@@ -0,0 +1,33 @@
...
+            std=prm["use_lifetime_std"],
+        )
+        stk["historic_in_use"].compute()
+
+    def compute_flows(self):
+        flw = self.flows
+        stk = self.stocks
+
+        flw["sysenv => use"][...] = stk["historic_in_use"].inflow
+        flw["use => sysenv"][...] = stk["historic_in_use"].outflow
```
JakobBD, line 33: Do you still use both simple and non-simple? 
If not, I would soon delete the non-used.

simson/cement/cement_definition.py
```diff
@@ -0,0 +1,201 @@
...
+        # fd.FlowDefinition(from_process="sysenv", to_process="raw_meal_preparation", dim_letters=("h", "r",)),
+        # fd.FlowDefinition(from_process="raw_meal_preparation", to_process="clinker_production", dim_letters=("h", "r",)),
+        # fd.FlowDefinition(from_process="sysenv", to_process="clinker_production", dim_letters=("h", "r",)),
+        # fd.FlowDefinition(from_process="clinker_production", to_process="cement_grinding", dim_letters=("h", "r",)),
+        # fd.FlowDefinition(from_process="sysenv", to_process="cement_grinding", dim_letters=("h", "r",)),
+        # fd.FlowDefinition(from_process="cement_grinding", to_process="concrete_production", dim_letters=("h", "r",)),
+        # fd.FlowDefinition(from_process="sysenv", to_process="concrete_production", dim_letters=("h", "r",)),
+        # fd.FlowDefinition(from_process="concrete_production", to_process="use", dim_letters=("h", "r", "s",)),
+        # fd.FlowDefinition(from_process="use", to_process="eol", dim_letters=("h", "r", "s",)),
+        # fd.FlowDefinition(from_process="eol", to_process="sysenv", dim_letters=("h", "r", "s",)),
```
JakobBD, line 54: Do you still need this, or can it be deleted? 
I would generally try to keep the amount of obsolete commented code to a minimum. 
It will still be stored in the git history, in case you want to recover it at some point. 
It's a rule of thumb, and there can be cases where it makes sense. 
Just keep in mind to delete it as soon as it's probably obsolete.

JakobBD (approved):
image

Looks great!! Nice work! 

I can't review every line of the code, but maybe it's okay if you are the only one who really knows it for now. At some point I'll have a closer read, but right now the priority is that it should be merged.

simson/cement/cement_definition.py
bennet21: Same here.

PR #40, opened 2025-03-20
Add fit boundaries 

### LINKED ISSUE #35: Better saturation level handling for extrapolation
saturation_level was added as a new attribute in Extrapolation and StockExtrapolation. A few further workarounds had to be employed to make everything work. Potentially this could be removed in the future, and instead it should be possible to prescribe ranges for every fit parameter, including saturation_level.

---
For the saturation_level, we could have a more general implementation, where we
- give each extrapolation class a list prm_names
- the parent class an argument fixed_prms which is a dict with name and value
- remove the FixedSatSigmoid class
- instead of n_prms, the parent class a "@"property n_prms which returns len(prm_names) - len(fixed_prms)
- this would require further changes, with an added function wrapper for func, which sets the fixed prms

### LINKED ISSUE #37: Delete unused historic cement mfa

bennet21:
# Features

Adds `Bound` class in data_transformations.py. `Bounds` takes the variable name, and upper and lower bounds as input. A static method `create_bounds_arr` is implemented that generates the bounds needed for `scipy.optimize.curve_fit` from a list of `Bound` object. For usage, if a variable should be constrained, a `Bound` object should be generated and passed as `bounds` to `StockExtrapolation` which are then forwarded to `Extrapolation`. Upper and lower bounds should be given with the same shape as `indep_fit_dim_letters`. If upper and lower bound are equal, the variable is effectively fixed.

# Limitations

- Setting very narrow, or even equal bounds might reduce convergence speed and ability. No issues were observed so far.
- Using the same bounds in `scipy.optimize.curve_fit` causes an error. A workaround has been implemented that makes the bounds slightly different.
- Improvements to a more flexible dimension of the bounds with automatic mapping would be a future feature update.

# Bug fixes

No bug fixes: the previous saturation level approach seems to not have caused any issues.

# Further Notes

- Closes #35 
- the behavior of `None` of `target_dim_letters` has been replaced with `"all"` to enhance consistency with `indep_fit_dim_letters`
- Remove unused historic cement MFA. Closes #37


simson/common/data_transformations.py
```diff
@@ -7,3 +9,60 @@ def broadcast_trailing_dimensions(array: np.ndarray, to_shape_of: np.ndarray) ->
...
+        self.upper_bound = np.array(upper_bound)
+
+    @staticmethod
+    def create_bounds_arr(
+        bounds_list: list["Bound"], all_prm_names: list[str], bound_shape: tuple
+    ) -> np.ndarray:
+        """Creates bounds array where each element is tuple of lower and upper bounds for each parameter."""
+
+        # Check bound shapes
+        if bounds_list:
```
JakobBD, line 34: I would make this more explicit
```suggestion
        if isinstance(bounds_list, iterable) and len(bounds_list) >= 1:
```

simson/common/data_transformations.py
```diff
@@ -7,3 +9,60 @@ def broadcast_trailing_dimensions(array: np.ndarray, to_shape_of: np.ndarray) ->
...
+
+            # Update bounds for parameters that have them
+            for bound in bounds_list:
+                pos = param_positions[bound.var_name]
+                lower_bounds[pos] = bound.lower_bound[index]
+                upper_bounds[pos] = bound.upper_bound[index]
+                if lower_bounds[pos] == upper_bounds[pos]:
+                    # avoid error in least_squares
+                    lower_bounds[pos] = lower_bounds[pos] * 0.999999
+                    upper_bounds[pos] = upper_bounds[pos] * 1.000001
```
JakobBD, line 64: doesn't work if both are zero

JakobBD: (or negative)

simson/common/data_transformations.py
```diff
@@ -7,3 +9,60 @@ def broadcast_trailing_dimensions(array: np.ndarray, to_shape_of: np.ndarray) ->
...
+        var_name: str,
+        lower_bound: Union[float, np.ndarray],
+        upper_bound: Union[float, np.ndarray],
+    ):
+        self.var_name = var_name
+        self.lower_bound = np.array(lower_bound)
+        self.upper_bound = np.array(upper_bound)
+
+    @staticmethod
+    def create_bounds_arr(
```
JakobBD, line 28: There is no reason to place this method inside the Bound class. It does not use its data, and is not called from its methods. Placing it here is therefore confusing, as one expects it to be connected to the class.

simson/common/data_extrapolations.py
```diff
@@ -79,6 +82,8 @@ def regress(self):
         target_shape = tuple([self.target_range.shape[i] for i in sorted(self.independent_dims)])
         regression = np.zeros_like(self.target_range)
         self.fit_prms = np.zeros(self.target_range.shape[1:] + (self.n_prms,))
+        bound_shape = tuple(self.target_range.shape[i] for i in self.independent_dims)
```
JakobBD, line 85: Is this necessarily the shape that the bounds have? 
Should that be ensured/checked somewhere?

simson/common/data_transformations.py
```diff
@@ -7,3 +9,60 @@ def broadcast_trailing_dimensions(array: np.ndarray, to_shape_of: np.ndarray) ->
...
+            raise ValueError("All bounds must have the same shape")
+
+        # Check for invalid parameter names
+        invalid_params = set(b.var_name for b in bounds_list) - set(all_prm_names)
+        if invalid_params:
+            raise ValueError(f"Unknown parameters in bounds: {invalid_params}")
+
+        bounds = np.empty(bound_shape, dtype=object)
+        param_positions = {name: i for i, name in enumerate(all_prm_names)}
+        for index in np.ndindex(bound_shape):
```
JakobBD, line 51: That loops seems quite inefficient. 
I think some array-based code would be better than nested loops and conditions.

JakobBD (commented):
Let's talk in person about bounds vs a solution where prescribed parameters are actually taken out of the least squares variables. I had a different intuition, but may change my mind once I look at this in more detail. 

Also, in general I have the feeling there should be more checks w.r.t. array shape, but I may be wrong. Will look into it. 

Sorry, this is more of an intermediate brain dump, rather than a review.,.

bennet21:
NEW: 
- updated code according to your suggestions
- class `BoundList` that takes a list of `Bound` objects and the target timensions. It handles upscaling these bounds to the target shape and has the `create_bound_arr` method to transform the `bound_list` into an array of bounds usable for scipy.least_squares.
- updated class `Bound`. Here, lower and upper bounds are now FlodymArrays such that their dimensionality can be easily adjusted.

bennet21:
## Ressource intensitiy of stock extrapolation: 
**Cement**: 13 %
running get_long_term_stock(): 0,04s
whole model (excluding export and plotting): 0,31s

**Steel**: 4,5 %
running get_long_term_stock(): 0,106
whole model (excluding export and plotting): 2,33s

JakobBD (approved):

PR #43, opened 2025-03-21
Add historic future generalized plot function

bennet21:
Note: Branch from [fit_boundaries](https://github.com/pik-piam/simson/tree/fit_boundaries). Requires [Flodym PR #96 ](https://github.com/pik-piam/flodym/pull/96).

# Features

Adds `plot_history_and_future` to `CommonDataExporter` (previously `CustomDataExporter`) which makes it possible to create history-future like plots for any variables. Additionally stock visualization was generalized into `visualize_use_stock` in  `CommonDataExporter` as it is used by both cement and steel.

# Notes
Generalization was used to plot production in cement. A similar approach could be used for steel production but is not yet implemented.


JakobBD:
okay if I only review this after #40 has been merged, to get a proper diff of the changes?

simson/common/common_export.py
```diff
@@ -0,0 +1,227 @@
...
+        elif self.cfg.plotting_engine == "pyplot":
+            return fde.PyplotArrayPlotter
+        else:
+            raise ValueError(f"Unknown plotting engine: {self.cfg.plotting_engine}")
+
+    def visualize_use_stock(self, mfa: fd.MFASystem, subplot_dim: str = None):
+        """Visualize the use stock. If subplot_dim is not None, a separate plot for each item in the given dimension is created. Otherwise, one accumulated plot is generated."""
+        per_capita = self.cfg.use_stock["per_capita"]
+
+        stock = mfa.stocks["in_use"].stock
```
JakobBD, line 83: maybe pass the stock as a parameter? might be called differently across MFAs

JakobBD (approved):
Looks good, thanks for this!

PR #49, opened 2025-04-18
adapt to flodym change where dims.shape is now a properyt instead of a method

JakobBD:
Needed to work with flodym after pik-piam/flodym#106. 

**Please wait with merging** until flodym PR is merged. Then also 
- create new flodym release
- bump flodym version in simson pyproject.toml accordingly.


bennet21 (approved):
very good

PR #70, opened 2025-10-01
Input Data from mrindustry

bennet21:
This PR is connected to https://github.com/pik-piam/mrindustry/pull/61.

# Changes
## MrindustryDataReader
To read data produced from mrindustry, a new reader is implemented. It reads all parameter as well as the region dimension. Other dimensions will have to be prescribed as before.
More detailed description:
- tgz file is extracted, saved in input_data in input folder (this only happens if data of that version isn't available yet)
- Dimensions are read from dimension folder, regions are inferred from unique elements in regionmapping 
- Parameters are read from cs4r files in input_data. Dimensions are inferred from its note. 

## Parameter Extension
For parameters, that only have historic values from mrindustry, they can be extended into the future using the parameter_extension file in common. Currently, only a constant method is implemented, but the framework is extendible to host other conceivable extensions. Extensions are instructed in the cfg model_switches (previously customization).


bennet21:
TODO:
- Add Force extraction switch

PR #72, opened 2025-10-02
Plastic MFA: split into historic/future MFA, introduce trade, improve stock extrapolation, plotting routines, make IAMC writer

leonieschweiger:
This PR introduces the following improvements/new features for the plastic MFA :

- the MFA is restructured into historic and future MFA
- trade is introduced at 4 stages: primary, intermediate, final plastic goods and waste
- stock extrapolation is now performed via regression over time and GDPpc, and per good and region separately. saturation level upper bounds are determined by a best-fit regression over all regions.
- plotting routines are improved to better assess the goodness of fit for the stock extrapolation and to plot individual flows
- the IAMC writer allows to export MFA outputs in IAMC data format


remind_mfa/common/stock_extrapolation.py
```diff
@@ -197,6 +212,12 @@ def gdp_regression(self):
         # transform back to total stocks
         self.stocks[...] = self.stocks_pc * self.pop
 
+    def gdp_time_regression(self, gdppc):
+        time = np.array(self.dims["t"].items)
+        # ToDo make this more flexible; ideally the parameter is calculated within this function or at least given as an argument
+        gdppc[...] = np.log10(gdppc[...]) * 70 + time[:, None, None]
```
JakobBD: Is there another log 10 in the regression function then? 
So should this rather be followed by a `10**x`, or directly  `gdppc * time**(1/70)`?

But: The above comment with renaming the parameter should also be reflected in the code, i.e. something like 

```
if regress_over == "time": 
    predictor = gdppc
elif regress_over == "weighted_sum": 
    predictor = weight_gdppcd * gdppc + weight_time * time
```

(this can mean that many occurences of the gdppc variable is renamed to predictor)

JakobBD (approved):
Great work, thanks!

PR #80, opened 2025-10-28
Trade improvements

leonieschweiger:
This PR does the following:
- remove unnecessary processes (imports and exports that only connect other processes with the respective trade markets)
- convert trade markets from stocks to simple processes and connect them to sysenv
- balance waste trade (which before was not balanced, but simply read from parameters)
- calculate material (and good) splits step by step for all flows (including trades) by backcalculating from the split in the use stock
- balance trades with material and good dimensions

### DIFF
--- a/remind_mfa/common/trade_extrapolation.py
+++ b/remind_mfa/common/trade_extrapolation.py
@@ -64,7 +64,7 @@ def extrapolate_trade(
     if balance_to is not None:
         future_trade.balance(to=balance_to)
 
-    # can't import more than what is demanded; can't export more than what is supplied. 
+    # can't import more than what is demanded; can't export more than what is supplied.
     future_first[...] = future_first.minimum(scaler)
     future_trade.balance(to="minimum")
 


remind_mfa/plastics/plastics_mfa_system.py
```diff
@@ -196,77 +167,94 @@ def compute_flows(self):
...
+            stk["in_use"].inflow,
+            "imports",
+            balance_to="hmean",
         )
 
-        flw["primary_market => primary_imports"][...] = (
-            trd["primary"].imports * material_element_split_noGood
+        flw["good_market => use"][...] = trd["final"].imports
+        flw["fabrication => good_market"][...] = trd["final"].exports
+        flw["sysenv => good_market"][...] = flw["good_market => use"][...]
```
JakobBD: The `[...]` on the right-hand side of an assignment doesn't do anything if you have it on the left-hand side.

JakobBD: there's quite some occurences of this, best to check the whole compute function...

PR #89, opened 2025-12-08
Mappings

JakobBD:
Move mappings to own data structure with inheritance & default handling, such that they can be used from both export and visualization routines. 

Needed to get back the more readable process names in the flow definition table.

Downside: Visualization mappings are not conveniently in the same file anymore. But this is inherently so, if we want to use them for data export as well. 

To discuss: 
For the file name mappings: Maybe we should extend the DimensionDefinition in REMIND-MFA by a file_name instead. That way, the file names are where one would look for them first?


leonieschweiger (approved):
Great stuff! Thanks :)
Yes, moving the file names to the dimension definitions would be more straight-forward. But I think it is also fine as it is right now. Could also be mentioned in the in-code docu eventually where to find this mapping.

PR #102, opened 2025-12-19
Docs polish2

leonieschweiger:
- remove assumptions from plastics docu for now
- make config an own category in docs
- add note regarding empty source fields in parameters table
- hot fix for preliminary plastic sources: there are two parameters that are currently constant - mechanical recycling yield and uncontrolled loss rate - which are hardcoded in mrmfa. I want to improve them soon (especially regarding granularity of mechanical recycling yield) and therefore don't want to invest work to include them in the madrat framework now. to make sure that their sources are still displayed, I quickly introduced a hotfix in merge_parameter_sources.py handing over custom source entries to be included in the parameters table. If you prefer to include something similar for your parameters instead of the note regarding expert guesses above the table, this can easily be added.


JakobBD (approved):
great, thanks!!

PR #106, opened 2025-12-19
remove import from docs script, do not allow extra or missing values

JakobBD:


JakobBD:
import caused docs action to fail

PR #123, opened 2026-02-18
Stock and trade extrapolation

JakobBD:
- Extrapolate stock using a fitting method, which modifies the parameters of a common regression based on a penalty function, which tries to minimize 0th and 1st derivative discontinuities at the past -> future transition  
- Extrapolate trade using a scaling with supply and demand, where one of the two is calculated with a fixed-point iteration cycle.


remind_mfa/common/common_model.py
leonieschweiger: The documentation string here is outdated - probably the whole can be deleted?

JakobBD: Thanks, I deleted it!

JakobBD:
Closing as outdated. Let's merge everything as part of #124

PR #136, opened 2026-04-10
Change Transition Behavior: Introduce PD-Controller

bennet21:
In this PR, we switch the blending method for historical and predicted stock values in the `remind_mfa/common/stock_extrapolation.py` module.

### Blending Preparation
`critically_damped_blend` stays at the center of the blending. Initial velocity `v0` is now found by an out-sourced function `_trend_slope`: 
* For long time windows, a propper polynomial fit ensures more stable slope results. Currently, degree of 1 (linear) is used by default (better results), but higher degree of 2 (quadratic) is possible.
* For time windows of length 1, a finite difference is applied (as before).
The size of the time window is lifetime-dependent (as before), but again in an outsourced function `_lifetime_dependent_n`:
* small bug fix (alpha now has correct normalization)

### Blending Procedure
The stock blending is now not added to the previously calculated stock prediction anymore. Instead, the transitioned stock is calculated by taking the intitial conditions as a base, now, instead of a static critically-damped system, the whole transitioned stock is now continuously forced in a spring-damper system towards the projection in terms of position and velocity - it is a PD controller. The new stock is calculated in `_integrate_transition`.
**Details:**
* The integration is performed using semi-implicit Euleur method.
* To guide the solution towards what the projection will be doing, and not what it is doing now, we apply a forward-looking velocity guidance. In the beginning, the reference velocity will be 5 timesteps ahead, which is linearly reduced to 1 timestep within half the `approaching_time`. This ensures a more direct response towards the end of the transitioning.
* With this dynamic approach, an eventual convergence of transition and projection is not guaranteed. To facilitate stable convergence, a *very* soft quintic blending is applied on top, which ensures a perfect match after 10x `approaching_time`. Shorter convergence periods lead to unrealistically fast convergence patterns in some cases (like China).
* The convergence criterion could also have been fulfilled by using a PID controller. However, the integrator part leads to oscillations as it aims to compensate areas. It is therefore not used.

### DIFF
--- a/remind_mfa/common/data_blending.py
+++ b/remind_mfa/common/data_blending.py
@@ -346,7 +346,6 @@ class CriticallyDampedBlender:
             n (int or np.ndarray): Smoothing window size. Either a scalar applied to all spatial
                 positions or an array matching the spatial shape of ``y``.
             idx (int): Time index at which to evaluate the slope (typically the last historical index).
-            deg (int): Maximum polynomial degree for the local fit. Defaults to 2.
             deg (int): Maximum polynomial degree for the local fit. Defaults to 1.
 
         Returns:
--- a/remind_mfa/common/data_blending.py
+++ b/remind_mfa/common/data_blending.py
@@ -238,7 +238,7 @@ class CriticallyDampedBlender:
         # overshooting. n_fwd decreases linearly from n_fwd_max to 1 over the first
         # n_ramp_steps, then remains 1 for the rest of the integration.
         n_fwd_max = 5
-        n_ramp_steps = int(approaching_time / 2)
+        n_ramp_steps = max(1, int((approaching_time / 2) / dt))
         n_fwd = np.maximum(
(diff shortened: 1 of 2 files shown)


bennet21:
In [e99cbe7](https://github.com/pik-piam/remind-mfa/pull/136/commits/e99cbe77a182635d4b31b470b7c1d2175776c0f3), the blender was moved to `common/data_blender.py`. Are you ok with that?

JakobBD:
Thanks for the great effort! It sounds complicated, but I believe that it was necessary for good results.
Could you provide before/after plots for the three materials? 
(This is standard practice in REMIND PR's, and now that we can all run all models conveniently, I'd like to establish it here as well). 

The two different blending functionalities in one file bother me a tiny bit, but I can see how they belong together somehow, so I'm happy to leave it there. Tomato, tomato.

JakobBD (approved):
Great stuff, thanks!

PR #137, opened 2026-04-10
add high value chemicals input

leonieschweiger:
This PR introduces a translation from plastics to HVC input, using input data from https://github.com/pik-piam/mrmfa/pull/54.
- The virgin process is renamed to polymerization. 
- The new polymerization yield parameter determines the total input into the polymerization process and losses.
- The total input into the polymerization process is split into HVC (w/o C4, because we do not account for C4 in REMIND), C4 and other reactants inputs (these are new processes) by two other new parameters (HVC and C4 input ratio). The remaining share of input is other reactants. 
- Chemical recycling now feeds into the new HVC input process. Hence, polymerization losses are now also accounted for here. For now, it is assumed that the carbon content of the HVC from chemical recycling is the same as of HVC going into polymerization, since the chemical recycling yield does not differentiate by material and element.
- For better understandability, the HVC feedstocks (fossil, ccu, daccu, bio) are renamed from virginXXX to feedXXX.

### DIFF
--- a/config/cement.yml
+++ b/config/cement.yml
@@ -3,7 +3,7 @@ model: 'cement'
 input:
   force_extract_tgz: False
   input_data_path: '../remind_mfa_data'
-  input_data_revision: '1.3.0'
+  input_data_revision: '1.4.0'
   region_mapping: 'h12'
   scenarios_path: 'config/scenarios'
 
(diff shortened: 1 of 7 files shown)


bennet21 (approved):
cool!

JakobBD (approved):
Great stuff, quite an improvement.

PR #138, opened 2026-04-28
One Big PR to Rule Cement

bennet21:
This PR changes a lot for the cement model. Accompanying mrmfa PR: https://github.com/pik-piam/mrmfa/pull/36.

# Updated Cement Base Model
### Add Dimension: Material Constituent (k)
Distinguishes between cement and non-cement product. As the stock is in terms of concrete/mortar, this dimension allows to easily get the cement content of the stock. So `cement_stock = concrete_stock * cement_ratio` becomes `cement_stock = concrete_stock["cement"]`.
### Remove Dimension: Product Application (a)
- this dimension distinguishes between different strength classes for concrete and application classes for mortar.
- it is now only used in the carbonation model, but not in the standard model anymore
### Streamline Parameters
Related to the adjusted Dimensions, the parameters where cleaned up.
- Cement ratio now replaces product density und product cement content (direct calculation in mrmfa). For simplicity, this is now only differentiates for region and material, not application anymore.
- Material application split combines product applications split, product_material_application_transform (only used in carbonatio model).
- remove stock saturation level (outdated)
### Parameter Extrapolation
Extension such that it works with all kinds of parameters:
    1. Parameters with 'h' dimension → extend to 't' dimension
    2. Parameters with 't' dimension → modify future values
    3. Parameters with no time dimension → add time dimension and apply transformation to future values
For this, it also needs to get the historic time dimension passed (adjusted in common model)
- added `SmoothScalingExtrapolation`, where we linearly scale by a factor. In the future, this could replace the cumbersome lifetime adjustment in the common model.
### Reworked cement mfa system future
- `compute_in_use_stock` can now take cement or concrete/mortar stock as input (requrired for bottom-up)
### Reworked cement mfa system historic(al)
- compute_cement_stock is now a staticmethod (it is used in parameter reconciliation)

## Minor changes
- Adjust cement cfg: carbonation is now binary option
- remove unused parameters
- fixed carbonation model: it finally works again yeeyy
- move approaching time in blend to stock extrapolation

# Introduce Bottom-up + Reconciliation
You may ignore the changes here for now, as it is not in a final version yet.
## Bottom-up MFA:
- new Dimensions, and parameters.
## Reconciliation
- The cement model now has an adjusted `run` function that expands the functionality of the common model: It calculates the bottom-up stock, does reconciles the stock through parameter adjustments, updates bottom-up and top-down stock, and merges them into a combined mfa (which optionally can become the new `self.future_mfa` if `self.cfg.model_switches.parameter_reconciliation.do_combine_mfas == TRUE`).
- most of the heavy lifting happens in `cement_parameter_reconciliation`


leonieschweiger (approved):
Great work, Bennet! :)
I only skimmed through most the cement stuff, but looks quite impressive!
Nice improvement of the ParameterExtrapolation - but should we rename it to ParameterTransformation accordingly? One thing that I find not ideal about the new SmoothScalingExtrapolation is that in the scenario.csv the parameter name is now ambigeous, because depending on the ParameterExtrapolation class is will either be the real parameter value in the target year, or a factor the parameter will be scaled with. Which of both is not clear from the scenario.csv, because the ParameterExtrapolation class is set in the config. I don't see a better solution at the moment though. What do you think @JakobBD?

remind_mfa/cement/cement_model.py
```diff
@@ -38,3 +41,162 @@ def modify_parameters(self):
...
         lifetime_std[...] = self.parameters["lifetime_mean"] * self.parameters["lifetime_rel_std"]
         self.parameters["lifetime_std"] = lifetime_std
+
+    def run(self):
+        self.original_parameters_hist = self.parameters.copy()
+        super().run()
+
+        if not self.cfg.model_switches.parameter_reconciliation.do_reconcile:
+            return
+
```
JakobBD, line 70:
```suggestion
        super().run()

        if self.cfg.model_switches.parameter_reconciliation.do_reconcile:
            self.run_with_reconciliation()

    def run_with_reconciliation(self):
        """Some doc"""    
        self.original_parameters_hist = self.parameters.copy()
```

Encapsulates the new stuff clearer (and makes it easier for the reader to turn around).

bennet21: Thanks for the suggestion. I implemented in a way very close to your suggestion. However, the original historical parameters still have to be saved before the `super().run()` which alters self.parameters.

PR #140, opened 2026-05-18
Modify trade extrapolation to scale linearly, but converge to global shares for large demand increases

JakobBD:
So far, a sub-linear trade scaling was used. 
This was only to fix issues with SSA (and other low-gdp regions)  having excessive trade otherwise. 
This PR fixes that, introducing linear trade scaling, but blending to the global import/export shares if demand increases a lot, such that historical trade shares for low-gdp regions play less of a role then.

### DIFF
--- a/remind_mfa/common/trade_extrapolation.py
+++ b/remind_mfa/common/trade_extrapolation.py
@@ -9,7 +9,6 @@ from remind_mfa.common.helpers import RemindMFABaseModel
 from remind_mfa.common.data_blending import blend
 
 
-
 class TradeExtrapolator(RemindMFABaseModel):
     """Predict future trade values by extrapolating the trade data using a given scaler."""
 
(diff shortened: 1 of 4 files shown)


leonieschweiger (approved):
Great work, thanks!

bennet21 (approved):

remind_mfa/common/trade_extrapolation.py
```diff
@@ -224,11 +218,11 @@ def balance(self):
...
-    @staticmethod
     def scaling(
+        self,
+        i0: fd.FlodymArray,
         d0: fd.FlodymArray,
         d: fd.FlodymArray,
-        alpha: float,
         reduced_linear: bool = True,
     ) -> fd.FlodymArray:
         """Apply scaling depending on relative change of scaler:
```
leonieschweiger, line 228: I just realized that the documentation for the scaling still needs to be updated here

JakobBD:
New commit (in a sec) will change alpha over (d/d0) as follows (red is old, green is new): 

image

New version has imo slightly improved trade (e.g. slightly higher exports for plastics, less weird cement SSA trend breaks), but no large changes.

remind_mfa/common/trade_extrapolation.py
JakobBD: thanks!

PR #142, opened 2026-05-21
Plastics fixes

leonieschweiger:
This PR does the following:
- Change scenario settings for plastics EoL rates: so far, quite outdated settings from Qianzhi had been used as targets for EoL rates (e.g. the global incineration rate target was lower than current incineration rates in several regions like JPN, which lead to incineration being replaced by landfilling); now, target rates are only set for recycling, and landfilling is phased out by 2050; the incineration rate is the remaining share that is not recycled or landfilled.
- Fix negative EoL flows by scaling down waste exports that were higher than waste collection for some polymers and regions.
- Modify sector split limits: so far, global sector split limits were assumed. However, this ignored regional differences which may be due to different consumption preference patterns that we would like to keep (e.g. JPN has much lower share in transportation than USA), but also due to different economic conditions (e.g. SSA has a much lower share in buildings and construction, but a higher share in packaging). For low-gdp regions, we cannot assume that the current split will be maintained in the future. Therefore, we use the historic split for highest gdp region, but blend to the global split if historic gdp is low in regions.
- Fix a weird jump in primary exports of Elastomers in JPN which originated from scale_second of TradeExtrapolator which uses future imports/max(historic imports,1) to scale exports with as a starting point of the iteration. The ratio is very large when historic imports are zero and the iteration could not recover from this. I now changed the initialization of the iteration to simply use historic exports for future exports.


remind_mfa/plastics/plastics_mfa_system.py
```diff
@@ -19,7 +19,7 @@ def compute(self, stock_projection: fd.FlodymArray, historic_trade: TradeSet):
         self.compute_flows(historic_trade)
         self.compute_other_stocks()
         self.check_mass_balance()
-        self.check_flows(raise_error=False)
+        self.check_flows(raise_error=False, verbose=True)
```
bennet21, line 22: Do you want to keep verbose on?

bennet21 (approved):
Exciting progress! Good to see some more plots added to plastics 😬

PR #143, opened 2026-05-21
Update Transition Blending Mechanism

bennet21:
This change has minor visible impact on the results (stock/demand)

The core functionality stays the same, however, the underlying integration method was improved and simplified:
1. In-integration blending is now more accurately called nudging. It is also now scaled quadraticly instead of quinticly for simplicity. Another option of increasing the damping parameter `k` quadratically over time leads to slight stock divergences and was discarded.
2. A boundary problem was fixed: gradients weren't calculated correctly at the last point.
3. The look-ahead velocity is now calculated continously using interpolation. This avoids potential artifacts.

Based on #138.


leonieschweiger (approved):
Great work Bennet, thanks! :)

Issue #87, opened 2025-12-01
Join export and plotting functions as much as possible


bennet21:
Move as much to CommonVisualization as possible.

Issue #82, opened 2025-11-13
Clean up plastics_export.py and move some functions to common_export.py


leonieschweiger:
- [ ] move `if self.do_export.iamc: self.write_iamc(mfa=model.mfa_future)` to common_export.py
- [ ] remove export_stock() function, as there is already an equivalent in common_export.py
- [ ] use  export_directory=dir_out from the parent class method for export functions in plastics_export.py
- [ ] export_mfa() in plastics_export.py should ideally take the same arguments as the parent function
- [ ] eventually move some plotting functions to common_export.py

PR #144, opened 2026-05-22
Clean up visualization

leonieschweiger:
Addressing issues #82 and #87 I moved all plotting functions used by multiple models to common_visualization.

### DIFF
--- a/config/plastics.yml
+++ b/config/plastics.yml
@@ -36,6 +36,9 @@ visualization:
     over_gdp: True
     accumulate_gdp: False
     per_capita: True
+  gdppc:
+    do_visualize: False
+    per_capita: False
   consumption:
     do_visualize: True
     per_capita: True
--- a/config/plastics.yml
+++ b/config/plastics.yml
@@ -36,7 +36,7 @@ visualization:
     over_gdp: True
     accumulate_gdp: False
     per_capita: True
-  gdppc:
+  gdp:
     do_visualize: False
     per_capita: False
   consumption:
--- a/config/plastics.yml
+++ b/config/plastics.yml
@@ -60,8 +60,10 @@ visualization:
         #- "intermediate_market"
         #- "good_market"
       exclude_flows: [ ]
-  flows:
+  sector_splits:
     do_visualize: True
+  flows:
+    do_visualize: False
 
 export:
   do_export: False
--- a/remind_mfa/cement/cement_visualization.py
+++ b/remind_mfa/cement/cement_visualization.py
@@ -126,26 +126,11 @@ class CementVisualizer(CommonVisualizer):
     def visualize_carbonation(self, mfa: fd.MFASystem):
         annual_uptake = mfa.stocks["carbonated_co2"].inflow
         cumulative_uptake = mfa.stocks["carbonated_co2"].stock
-        linecolor_dimletter = "Carbonation Location"
-        plot_letters = ["t", "c"]
-        other_dimletters = tuple(
-            letter for letter in annual_uptake.dims.letters if letter not in plot_letters
+        linecolor_dim = "Carbonation Location"
+        self.visualize_flow_stacked(
+            mfa=mfa,
+            flow=annual_uptake,
+            name="Annual CO2 uptake from carbonation",
+            linecolor_dim=linecolor_dim,
+            regional=False,
         )
-        annual_uptake = annual_uptake.sum_over(other_dimletters)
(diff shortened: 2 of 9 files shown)


config/plastics.yml
```diff
@@ -36,6 +36,9 @@ visualization:
     over_gdp: True
     accumulate_gdp: False
     per_capita: True
+  gdppc:
```
bennet21: I would call this gdp as pc is an option.

remind_mfa/cement/cement_visualization.py
```diff
@@ -183,81 +123,6 @@ def visualize_global_stock_by_type(
...
-        extrapolation_name = "_extrapolation" if show_extrapolation else ""
-        future_name = "_projection" if show_future else "_historic"
-        self.plot_and_save_figure(
-            ap,
-            f"cement_stocks{extrapolation_name}{future_name}.png",
-            do_plot=False,
-        )
-
     def visualize_carbonation(self, mfa: fd.MFASystem):
         annual_uptake = mfa.stocks["carbonated_co2"].inflow
```
bennet21, line 71: I could use your new `visualize_flow_stacked` here, too! :)

remind_mfa/common/common_config.py
```diff
@@ -106,6 +106,11 @@ class ConsumptionVisualizationCfg(BaseVisualizationCfg):
     """Whether to visualize consumption per capita."""
 
 
+class GDPVisualizationCfg(BaseVisualizationCfg):
```
bennet21, line 109: here, you chose a better name imo :)

remind_mfa/common/common_visualization.py
```diff
@@ -146,11 +156,11 @@ def visualize_use_stock(
         # Adjust x-axis
         if self.cfg.use_stock.over_gdp:
             if self.cfg.plotting_engine == "plotly":
-                fig.update_xaxes(type="log", range=[3, 5])
```
bennet21, line 149: After removing the x range/lim, will the x-axis still have the same range for all subplots (which is what we want, right?). Maybe we can use `shared_xaxes=True` (plotly) and `sharex=True` (matplotlib) instead?

remind_mfa/common/common_visualization.py
```diff
@@ -380,3 +387,221 @@ def visualize_trade(
...
+            xlabel="Year",
+            ylabel="Sector Splits [%]",
+            display_names=self.display_names.dct,
+            title=f"Product demand sector splits ({name_str})",
+            chart_type="area",
+        )
+
+        self.plot_and_save_figure(ap_sector_splits, f"sector_splits_{name_str}.png")
+
+    def visualize_flow(
```
bennet21: why can this not be used for anything, including stock. Could this be `visualize_fdarr` instead? Or am I missing something?

remind_mfa/common/common_visualization.py
```diff
@@ -380,3 +387,221 @@ def visualize_trade(
...
+            linecolor_dim=linecolor_dim,
+            x_label="Year",
+            y_label=f"{name} [t]",
+            title=f"{name} {pc_str} {regional_tag}",
+            line_label=name if linecolor_dim is None else None,
+        )
+
+        self.plot_and_save_figure(ap_flow, f"{name}_{pc_str}_{regional_tag}.png", do_plot=False)
+
+    def visualize_flow_stacked(
```
bennet21: same here :)

remind_mfa/steel/steel_config.py
```diff
@@ -1,18 +1,9 @@
...
-    per_capita: bool = False
-    """Whether to visualize stock per capita."""
-
-
 class SteelVisualizationCfg(VisualizationCfg):
     scrap_demand_supply: BaseVisualizationCfg
     """Visualization configuration for scrap demand and supply."""
-    sector_splits: BaseVisualizationCfg
-    """Visualization configuration for sector splits."""
-    gdppc: GDPVisualizationCfg
```
bennet21, line 14: ah now i see where the gdppc comes from. I still think we should rename to gdp

bennet21 (approved):
This is very helpful work, thanks Leonie!

remind_mfa/common/common_visualization.py
leonieschweiger: They already have the same range on the x-axis. I think this range was chosen to zoom into the space of interest, however this was a problem when per capita=False where total stocks are plotted over total GDP, which is why I removed it.

PR #146, opened 2026-06-01
feat: plot parameters

bennet21:
[Vibe-coded]
Creates a PDF with plots of all parameters across all of their dimensions. Historical and Future parameters are plotted separately as sanity check (they should ofc overlap).

### DIFF
--- a/remind_mfa/common/common_config.py
+++ b/remind_mfa/common/common_config.py
@@ -69,6 +69,11 @@ class BaseExportCfg(RemindMFABaseModel):
     """Path to export folder for this entity"""
 
 
+class ParameterPlotsExportCfg(BaseExportCfg):
+    do_export: bool = False
+    """Whether to export the parameter sanity-check PDF."""
+
+
 class ExportCfg(BaseExportCfg):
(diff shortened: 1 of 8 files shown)


remind_mfa/common/common_config.py
```diff
@@ -69,6 +69,11 @@ class BaseExportCfg(RemindMFABaseModel):
     """Path to export folder for this entity"""
 
 
+class ParameterPlotsExportCfg(BaseExportCfg):
+    do_export: bool = False
```
JakobBD, line 73: I think I'd rather have it added to the config files with False instead of making a new subclass here. 
Alternatively, you could rename the subclass to `OptionalExportCfg` to make it more general.

JakobBD (commented):
Not sure what to do with this. 
It's nice added functionality, and the vibe coded bit are mostly confined to one file, but still I think it would be nice to refactor this before merging: 
- split up the main function a bit
- comment your code
- It seems like a lot of duplicated functionality with the ArrayPlotter. Maybe use that instead to shorten the code?

PR #147, opened 2026-06-02
Allow trade of recycled plastic polymers

leonieschweiger:
Previously, recycled plastics were directly going to plastics fabrication, implying that all recycled plastics are used within a region. To account for the fact that also recycled granulate is traded, I now route the mechanically recycled plastics flow to primary market. This lead to the problem that more recycled plastics were available than the plastics supply resulting from the primary trade extrapolation for some polymers within some regions. Assuming that this oversupply of recycled plastics will lead to either increased exports or decreased imports, I corrected the respective primary trade (using an iteration to ensure global trade balancing).


bennet21 (approved):
good stuff

PR #149, opened 2026-07-06
chore: use uv to manage venv

tobiasdiez:
Use `uv` in the readme (and to build the docs), instead of plain pip.

Also change the build backend to uv (since its a bit faster, and still actively maintained which you cannot really say about setuptools).

Background: If I understood Tonn correctly (to be double-checked tomorrow), then uv is the easiest way to create a reproducible venv on the cluster. But independently of the use case, uv is the modern replacement of pip.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ config/
 *.pyo
 .idea
 .vscode
-uv.lock
 remind_mfa.egg-info
 site/
 .github/copilot-instructions.md
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -19,4 +19,7 @@ mkdocs:
 # See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
 python:
    install:
-     - requirements: docs/requirements.txt
+     - method: uv
+       command: sync
+       groups:
+         - docs
--- a/README.md
+++ b/README.md
@@ -7,22 +7,33 @@ REMIND-MFA has global coverage. Per default, it runs in 21 world regions. Howeve
 
 ## Installation
 
-REMIND-MFA dependencies are managed with [pip](https://pypi.org/project/pip/).
+REMIND-MFA dependencies are managed with [uv](https://docs.astral.sh/uv/).
 
 To install, clone the repository and run
 
 ```
-python -m pip install -r pyproject.toml
+uv sync
 ```
 
 from the repository's main directory.
+This installs the project together with some development dependencies.
+For a minimal installation, you can run `uv sync --no-dev` instead.
+
+For development, it may be convenient to check out the `flodym` repository in the same parent directory as `remind-mfa` and uncomment the `tool.uv.sources` section in `pyproject.toml` to point to the local `flodym` repository.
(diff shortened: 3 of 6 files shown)


JakobBD:
Thanks for the upgrade! 
I'm happy to switch to a cleaner workflow. 
We haven't worked wit uv lock files so far, but not for very good reasons, so I'm happy to add them. 

It may be good for us to have a short intro to uv workflows, either written or orally. If you write it down, I think REMIND-MFA is the best place rather than flodym. I'm sure, there's also existing tutorials online? Should just be short and limited to the bits we need regularly.
We've obviously worked with it before, but I have some questions on standard practices (and on VSCode integration in particular), and I'm certain the others do as well.

As for the editable flodym: It has eased our workflow in the past to have it editable by default, and we were the only REMIND-MFA users. I see, of course, that this changes soon, so it's only a matter of time until we have to change it. So I'm also okay with doing it now.

tobiasdiez:
> It may be good for us to have a short intro to uv workflows, either written or orally. If you write it down, I think REMIND-MFA is the best place rather than flodym. I'm sure, there's also existing tutorials online? Should just be short and limited to the bits we need regularly. We've obviously worked with it before, but I have some questions on standard practices (and on VSCode integration in particular), and I'm certain the others do as well.

I searched a bit online, and was surprised that I couldn't find a good general guide that focuses on the basics of _using_ uv to setup a venv and running commands. As probably more people would be profiting from such a guide, I've added one in https://github.com/scientific-python/cookie/pull/829 to the scientific python website. Rendered version can be found at https://scientific-python-cookie--829.org.readthedocs.build/en/829/guides/uv/. If you have any suggestions, please add them either here or in that PR (or tell me in our next call). 

> 
> As for the editable flodym: It has eased our workflow in the past to have it editable by default, and we were the only REMIND-MFA users. I see, of course, that this changes soon, so it's only a matter of time until we have to change it. So I'm also okay with doing it now.

Sadly, I don't have a good solution for this at the moment. There are a couple of open issues in uv, some with partial workarounds, eg https://github.com/astral-sh/uv/issues/7945 and https://github.com/astral-sh/uv/issues/9258#issuecomment-2824992171. In my limited testing, it sufficed to uncomment the line once, then run `uv sync`, then restore the `pyproject.toml`. Further usages of `uv run` kept using the editable install of flodym. That was good enough for me so far.

JakobBD (approved):
Thanks for the tutorial! 
Only thing "missing" from my point of view would be how to integrate that with VSCode (debugger and tests), though this tutorial is probably not the right place for that.

As I said, let's try this solution with the editable flodym. If it's too much of an annoyance in our daily workflow (I don't think it will be), we can still go the dirty way and uncomment the respective lines by default.

tobiasdiez:
> Thanks for the tutorial! Only thing "missing" from my point of view would be how to integrate that with VSCode (debugger and tests), though this tutorial is probably not the right place for that.

Thanks for the feedback. For vscode, it should correctly discover everything after you run `uv sync`.

> As I said, let's try this solution with the editable flodym. If it's too much of an annoyance in our daily workflow (I don't think it will be), we can still go the dirty way and uncomment the respective lines by default.

:+1:

PR #150, opened 2026-07-06
feat: load environment variables from a `.env` file in `run_remind_mfa.py`

tobiasdiez:
Add the capatibility to load env variables from a `.env` in the root of the repo. This is usally more convienient than manually setting env variables.

(Depending on wether you consider `run_remind_mfa` to be part of the package, one still needs to add python-dotenv as a package dependency.)


JakobBD:
That's a cool functionality! 
`MADRAT_OUTPUTFOLDER` is also used by madrat. That way, if mrmfa is run on the same system as the mfa, the folders will automatically coincide if the env variable is the same for both mrmfa and mfa, avoiding a copying step. This may be a reason to discourage setting it package-locally?
That's no objection from my side though, just a remark.  

And yes, I would add dotenv it to the package depenedencies then.

tobiasdiez:
> MADRAT_OUTPUTFOLDER is also used by madrat. 

Right, on the cluster you would just set `MADRAT_OUTPUTFOLDER` and then it works nicely. But I wanted to have a version that also works offline (just take a train through Brandenburg and Sachsen Anhalt, and you know why). So, I have a project-local `madrat` folder which contains the necessary files copied from the cluster (also has the advantage that I don't accidentally modify important files on the cluster). I have a short script that downloads the right file from the cluster - happy to share this one.

> And yes, I would add dotenv it to the package depenedencies then.

done

JakobBD (approved):
Preventing files on the cluster from being accidentally modified?? Preposterous, that's against our culture!

Looks great, happy to merge! 

NB: When merging with the revamp config PR, I would move `load_dotenv` to the `main()` routine.

PR #151, opened 2026-07-06
Fix capitalization of 'BOF' in parameters

tobiasdiez:
The file is called `st_scrap_in_BOF_rate.cs4r` with a capital `BOF`, which lead to errors as it couldn't be found (perhaps a linux/windows issue).

Thus used the capitalized version everywhere (where I could find ` scrap_in_bof_rate`). The alternative would be renaming the file to lower case.


JakobBD (approved):
Interesting that Windows seems to be case insensitive when looking for files? 

I'd prefer to change the file name (no good reason for capitalization here). But it requires to have the R packages set up. 

If you want to merge it now to have a working version, I'm fine with it! 
We can revert the commits once we changed mrmrfa (probably never :-D )

tobiasdiez:
Proper fix is at https://github.com/pik-piam/mrmfa/pull/64

PR #152, opened 2026-07-06
chore: improve path validation and error messages in `CommonDataReader`

tobiasdiez:
... to get earlier and more detailed errors in case something is configured wrongly (I know, who would do such a thing :smile:)


JakobBD (approved):
Great, thanks!

PR #153, opened 2026-07-06
Fix path to region mapping file

tobiasdiez:
I don't have a regionmapping.csv file, but a `regionmappingH12.csv`.


JakobBD:
I would hope this becomes obsolete with the new input data version in #154.

tobiasdiez:
Indeed!

PR #154, opened 2026-07-06
increment input data version to 1.6.0

bennet21:


tobiasdiez (approved):
This seems to work for me. But I don't (yet) know what other consequences this change has (and what the normal procedure for the update is), so I let @JakobBD give the final green light.

bennet21:
> This seems to work for me. But I don't (yet) know what other consequences this change has (and what the normal procedure for the update is), so I let @JakobBD give the final green light.

It only has implications for the reconciliation which I will fix in a subsequent PR. So this is definitely good for the moment.

PR #155, opened 2026-07-13
feat: add SSP3-SSP5 and CE scenarios

bennet21:
This PR contains updates mainly focused on the Bottom-up approach. However, some changes carry over to the top-down part as well, which is why I think it is useful to merge it.
Let me know
1. If you dislike having constant (but everchanging) updates to the bottom-up part in main.
2. What you think of the new scenario setup and the way drivers are now handled?

### Scenario Inheritance
Scenarios now inherit from a bottom-up base and a top-down base csv that contain standard settings. These base settings should be (but are not enforced to be) orthorgonal to each other. Essentially all settings that were previously found in SSP2 are now found in the top-down base. What do you think of this setup?
### New Scenarios
I added some new scenarios (SSP3-5 and SSP1-2 CE) with a first estimate of their top-down parameter adjustments (Saturation level, lifetime).
### Generalizing driver scenario S
With input data gen 1.6.0, floorspace became driver scenario dependent. This required a generalization of how this dimension is handled: `gdp_pop_scen` is now called `driver_scen` and every parameter that contains `S` will be sliced accoriding to the selected scenario (see `common_model`).
### Generalized make_mfa (in `common_model`)
This is needed to make a bottom-up mfa in the `cement_model`. Everything stays backwards-compatible.
### Added extrapolation class
This will be replaced by #157, so the following explanation can be skipped:
`ConstrainedSplitExtrapolation` is intended for extrapolations of shares that sum up to one over 1 dimension. For more than 2-item dimensions you can further provide a receiver item that swallows the change from the targeted item. I tried to make this class a bit flexible: the dimension and the blend type can be defined via the class name. Lmk if you have a better idea.

### Separated Bottom-up MFA system
Buttom-up parameters like building structure splits are now acting on the *inflow* to the stock, instead of the stock itself. This makes much more sense as it is what societies could change, and therefore what should be reflected in scenarios. To accomplish this, while all the splits are refering to the stock, we did the following. For the historical period, we assume a constant split of the inflow, which is equal to the split of the stock. As long as the lifetimes are not dependent on the dimensions of the split, this ensures that the stock also follows this split, which is what we have data on. Going into the future, we can now gradually change these splits according to scenario assumptions. All of this requires a bit of back-and-forth calculations, which are now found in `cement_mfa_system_bottom_up.py`.


config/scenarios/SSP1_CE.csv
```diff
@@ -0,0 +1,3 @@
+git parameter,models,value,extra:year,index:Region,description
```
leonieschweiger:
```suggestion
parameter,models,value,extra:year,index:Region,description
```

leonieschweiger (approved):
Looks good to me! 
With the new scenario inheritance, the base scenarios alone cannot run because they have no driver_scen, right? Another option would be to have SSP2_bottom-up and SSP2_top-down, then every scenario would be a "real" scenario, but I see that the way you implemented it is a bit nicer to read.

bennet21:
> Looks good to me! With the new scenario inheritance, the base scenarios alone cannot run because they have no driver_scen, right? Another option would be to have SSP2_bottom-up and SSP2_top-down, then every scenario would be a "real" scenario, but I see that the way you implemented it is a bit nicer to read.

Yes you have got a point and this might actually be a killer argument for this approach. Curious to hear the opinon of the others. Your suggestion would solve this. SSP2 would stand out in this case, but that might even be desired as it is our baseline scenario.

bennet21:
I also removed the SSP1_lowEU scenario for steel - I assumed it is not really used anymore, right @JakobBD?

bennet21:
This PR makes some adjustements to `parameter_extrapolation` - a file that kept getting uglier and uglier. I tried cleaning it up in #157, so you can skip over the `parameter_extrapolation` file here when reviewing the code - and give me your opinion on the new structure in the new PR :)

remind_mfa/cement/cement_mfa_system_bottom_up.py
```diff
@@ -0,0 +1,137 @@
...
+        6. Blend td into bu stock where bu is available, use td everywhere else.
+        7. Compute the complete MFA with the blended stock and historic trade.
+        """
+        if scale:
+            raise NotImplementedError("Scaling not implemented for bottom-up system.")
+
+        prm = self.parameters
+        stk = self.stocks
+
+        # ------------- Compute BU stock -------------
```
tobiasdiez: Usually, if you need comment headers to structure the code, it's better to extract the parts to separate methods. Here, `compute_floorspace`, `compute_bu` etc.

remind_mfa/cement/cement_mfa_system_bottom_up.py
```diff
@@ -0,0 +1,137 @@
...
+    def compute_bottom_up_concrete_stock(
+        floorspace: fd.FlodymArray, prm: dict[str, fd.FlodymArray]
+    ) -> fd.FlodymArray:
+        """Concrete in-use stock from a floorspace quantity, resolved by
+        building function (f) and structure (b).
+        """
+        function_split = prm["function_buildings_split"]
+        structure_split = prm["structure_buildings_split"]
+        function_split = function_split.get_shares_over(("f",))
+        structure_split = structure_split.get_shares_over(("b",))
```
tobiasdiez: Combine to `function_split = prm["function_buildings_split"].get_shares_over(("f",))`?

remind_mfa/common/parameter_extrapolation.py
```diff
@@ -306,17 +304,183 @@ def apply_prm_extrapolation(
...
-                raise ValueError(f"scenario_parameters required for {extrapolation_class.__name__}")
-            return extrapolation_class(scenario_parameters)
+                raise ValueError(f"scenario_parameters required for {cls.__name__}")
+            return cls(scenario_parameters)
         else:
-            return extrapolation_class()
+            return cls()
+
+
+class ConstrainedSplitExtrapolation(ScenarioExtrapolation):
```
tobiasdiez, line 322: I would suggest to add a few tests for this class, especially since it may be also used in other contexts.

tobiasdiez (changes requested):
There is a lot going on this PR that I don't understand, but code-wise it looks mostly good to me. Just a couple of minor suggestions from my side.

bennet21:
From @leonieschweiger [comment](https://github.com/pik-piam/remind-mfa/pull/157#discussion_r3643446207) in #157:
> For me it is not clear why you would not put this into base_top-down - maybe having just an SSP2 scenario as our baseline and all others inherit from there is less ambiguous than having this "base_top-down". But let's discuss this in https://github.com/pik-piam/remind-mfa/pull/155

I am happy to change this if others feel the same. The way it is implemented now, scenarios are more symmetric and their drivers are more transparent. But I totally see your point - this is a design question. Any other opinions that can tip the balance? 😬 @JakobBD @tobiasdiez 

This is also connected to @leonieschweiger's earlier comment: https://github.com/pik-piam/remind-mfa/pull/155#pullrequestreview-4703971014

remind_mfa/common/parameter_extrapolation.py
bennet21: This specific class is overwritten in #157 anyways. But I think your point holds for the entire `ParameterExtrapolation` class there :)

tobiasdiez:
Is there a use case for having the two different bases `base_bottom-up` and `base_top-down`? The former seems to be the base for all real scenarios, while the later is only used for `base_bottom-up`. If not, you could just have one `base` and the SSP scenarios are overwriting this.

bennet21:
I meant to differentiate between BU and TD settings to reduce confusion, as atm only TD widely used. But yes, also one singular `base` is possible. Let's discuss in the JF :)

JakobBD:
I agree with Tobias. I've suggested before to merge the two files.

edit: 
you could have sections in the file, separated by titles as comments.
Would need to make sure the parser can handle it.

bennet21:
Closed in favor of #157

PR #156, opened 2026-07-16
feat: revamp config loading

tobiasdiez:
Migrate from seperate per-model config files to more comprehensive config files that can be layered.

In particular:
- Replaced YAML config files with a single TOML configuration that has shared `base` settings plus per-material overrides. (No strong opinion about using toml, but its the 'pythonic' config format and Python has bultin support for reading such files)
- Added a config loader that deep-merges across multiple config layers, and then afterwards folds the model section over the base config.
- Slightly reworked the CLI runner to use Typer in order to support repeated `--config` options for layered overrides and `--model` to specify what model to run.
Using typer it was then also easly to add an interactive prompt when no arguments are provided to make Bennet happy.

Fixes https://github.com/pik-piam/industry_issues/issues/11.

Possible follow-ups:
- The paths in the `visualization` and `exports` section are material-specific but actually follow the same pattern. We could remove those seperate config options by either using placeholders (eg `figures_path = "data/%model%/output/figures`) or use a non-configurable convention about the folder structure.
 - Make it possible to run multiple material models (at the moment either all or a single one can be selected)


pyproject.toml
```diff
@@ -13,12 +13,12 @@ authors = [
     {name = "Merlin Hosak"},
 ]
 
-requires-python = ">= 3.10"
+requires-python = ">= 3.12"
```
tobiasdiez, line 16: Python 3.12 is needed for built-in `toml` loader and using the `type` alias keyword. There are other workarounds, but the Python upgrade was the easiest. The change is aligned with https://scientific-python.org/specs/spec-0000/, but not sure what Python version convention you guys want to follow.

tobiasdiez:
This is a draft for now as I still want to double-check that the config is effectively the same and the model runs result in the same output.

Early feedback, especially about the new `default.toml` config is nonetheless appreciated.

leonieschweiger:
Looks good to me, thanks Tobias! 
We could probably declutter the default.toml a bit by agreeing on a visualization and export set for all materials that we would like to plot/export regularly in the team, then the material-specific sections could be shortened and this would be the standard config to run in order to check changes for all materials. In addition, everyone could have their own .toml with user-specific settings (e.g. for me, plotting plastics in more detail).

remind_mfa/common/helpers.py
```diff
@@ -1,13 +1,39 @@
...
+    from remind_mfa.common.common_model import CommonModel
+
 
 class ModelNames(str, Enum):
     PLASTICS = "plastics"
     STEEL = "steel"
     CEMENT = "cement"
 
 
+def get_model_classes():
```
JakobBD: For lazy loading, we could have this function take the model name as an argument, and only import the needed model? Maybe in a match/case?

remind_mfa_cement.py
```diff
@@ -1,7 +1,7 @@
 # %%
+from remind_mfa.common.helpers import ModelNames
```
JakobBD: While we're at it, this file could be deleted, along with those for the other materials.

remind_mfa.py
```diff
@@ -31,30 +21,67 @@ def configure_logger():
...
-def read_model_config(filename):
-    with open(filename, "r") as stream:
-        data = yaml.safe_load(stream)
-    return {k: v for k, v in data.items()}
+def run_remind_mfa(config_names: list[str], selection: ModelSelection) -> None:
+    configure_logger()
+    selected_models = list(ModelNames) if selection == "all" else [selection]
+
+    for model in selected_models:
+        model_config = load_config(config_names, model)
```
JakobBD, line 27: There's currently no option to specify a config path other than `cwd/config`, and no option to give paths to individual config files. 

I find that behaviour fine when prompting for config files, but at least I would find it more intuituve if 

`uv run run_remind_mfa.py --config config/default.yaml` would not return an error.

Personally, I would change the behaviour such that `--config` always expects the full (relative) path, while prompting looks in the `config` folder (could even be relative to the repo instead of cwd). 

I see the downside of having to give the full path several times for several config files, but the only use case for that right now is running on the cluster for input data generation.

remind_mfa.py
```diff
@@ -1,25 +1,15 @@
 import logging
-import yaml
-import sys
+from typing import Annotated, Literal
```
JakobBD, line 2: While we're at it, rename file to `remind_mfa.py`? wdyt?

remind_mfa.py
```diff
@@ -31,30 +21,67 @@ def configure_logger():
...
+    entered_names = typer.prompt(
+        f"Configs (comma-separated, available: {choices})", default="default"
+    )
+    return [name.strip() for name in entered_names.split(",") if name.strip()]
 
 
-def init_model(cfg: dict) -> CommonModel:
-    """Choose MFA subclass and return an initialized instance."""
+@app.command()
+def main(
```
JakobBD, line 59: To make that more convenient, you could add a (boolean) `--remind-input` option, which sets all the others (config files, materials, later also scenarios).

run_remind_mfa.py
```diff
@@ -31,30 +21,67 @@ def configure_logger():
     )
 
 
-def read_model_config(filename):
-    with open(filename, "r") as stream:
-        data = yaml.safe_load(stream)
-    return {k: v for k, v in data.items()}
+def run_remind_mfa(config_names: list[str], selection: ModelSelection) -> None:
+    configure_logger()
+    selected_models = list(ModelNames) if selection == "all" else [selection]
```
JakobBD: I would move these two lines to `main()`, and pass `selected_models` to `run_remind_mfa`, only calling it `models` there.

config/default.toml
```diff
@@ -0,0 +1,116 @@
+[base.input]
```
JakobBD: As discussed, I would personally prefer switching back to yaml - my apologies for the miscommunication and the resulting back and forth, that was my fault! 

For our deep hierarchy and nesting (I count up to 6 levels), I find the yaml indentation much easier to read than the toml dot notation. For me that outweighs toml files being somewhat more pythonic.  

I don't want to decide this alone. @leonieschweiger and @bennet21 - What do you think?

pyproject.toml
JakobBD: Fine for me!

JakobBD (commented):
This is great stuff, so helpful!!! 
Thanks, it's greatly appreciated. 

I have commented the point on yaml (Leonie and Bennet, please comment!), and a few nit-picky points of personal preference, and a few suggestions for further improvements.

JakobBD:
> We could probably declutter the default.toml a bit by agreeing on a visualization and export set for all materials that we would like to plot/export regularly in the team

I agree! @leonieschweiger  could you make a suggestion?

> In addition, everyone could have their own .toml with user-specific settings (e.g. for me, plotting plastics in more detail).

I like the idea! We could add the folder to .gitignore, and except `default.yaml` and `cluster.yaml`?

config/default.toml
leonieschweiger: I also find the yaml indentation more clear and intuitive. If the advantages of toml are not big (which I cannot judge myself), I would also be in favour of keeping yaml, but I think I could adjust to both :)

tobiasdiez: @JakobBD We discussed this a bit. There is no strong preference from our side to either option. Yaml has the advantage of being familiar, toml that it _optionally_ allows to extract groups of keys to a table. We discussed that eg `sankey` under visualization might be better as a table/indented subgroup, while the single-value options like `extrapolation.do_visualize` feel cleaner in the toml dot notation than with forced indention in yaml.

toml does allow indentation, and that would be our proposal (see last commit) but going back to yaml is definitely also a fine option for us.

config/default.toml
```diff
@@ -0,0 +1,120 @@
...
+        parameter_extrapolation.emission_capture_rate = "ConstantExtrapolation"
+        parameter_extrapolation.material_shares_use_inflow = "ConstantExtrapolation"
+
+    [plastics.visualization]
+        figures_path = "data/plastics/output/figures"
+        consumption.do_visualize = true
+        consumption.per_capita = true
+        trade.do_visualize = true
+        sankey.do_visualize = true
+        sankey.plotter_args.slice_dict.e = "C"
```
JakobBD:
```suggestion
        sankey.plotter_args.slice_dict = {"t": 2050, "e": "C"}
```

config/default.toml
```diff
@@ -0,0 +1,120 @@
...
+        consumption.do_visualize = true
+        eol_stock.do_visualize = false
+        eol_stock.over_gdp = false
+        eol_stock.per_capita = false
+        prod_clinker.do_visualize = false
+        prod_cement.do_visualize = true
+        prod_product.do_visualize = false
+        sankey.plotter_args.split_flows_by = "Stock Type"
+        sankey.plotter_args.color_scheme = "blueish"
+        sankey.plotter_args.flow_color = "hsl(230,20,70)"
```
JakobBD, line 92:
```suggestion
        sankey.plotter_args = {
            "split_flows_by": "Stock Type",
            "color_scheme": "blueish",
            "flow_color: "hsl(230,20,70)",
        }
```

JakobBD: I suppose toml doesn't support newlines in dicts? then maybe discard it...

config/default.toml
```diff
@@ -0,0 +1,120 @@
...
+        use_stock.per_capita = true
+        use_stock.accumulate_gdp = false
+        gdp.do_visualize = false
+        gdp.per_capita = false
+        production.do_visualize = false
+        trade.do_visualize = false
+        consumption.do_visualize = false
+        consumption.per_capita = false
+        sankey.do_visualize = false
+        sankey.plotter_args.slice_dict.t = 2050
```
JakobBD:
```suggestion
        sankey.plotter_args.slice_dict = {"t": 2050}
```

remind_mfa/common/helpers.py
tobiasdiez: Makes a lot of sense, and the calls are also easier to read. Thanks for the suggestion!

remind_mfa.py
tobiasdiez: I've now changed so that both `--config config/defaut.toml` as well as `--config default` work. Is that okay with you, or do you want to enforce the relative path version?

remind_mfa.py
tobiasdiez: That's an interesting idea. If it's okay with you, I would it leave for a potential follow-up after I get a bit more experience on how often and in which way I will run this script. I could also imagine that you guys would like to have custom shortcuts for different settings so maybe a more generic solution would be worthwhile to implement.

config/default.toml
```diff
@@ -0,0 +1,110 @@
...
+        sector_splits.do_visualize = true
+        flows.do_visualize = false
+
+    [plastics.export]
+        do_export = false
+        path = "data/plastics/output/export"
+        csv.do_export = true
+        iamc.do_export = true
+        assumptions.do_export = true
+        docs.do_export = true
```
leonieschweiger:
```suggestion
        path = "data/plastics/output/export"
```

config/default.toml
```diff
@@ -0,0 +1,110 @@
...
+        prod_product.do_visualize = false
+        sankey.plotter_args.split_flows_by = "Stock Type"
+        sankey.plotter_args.color_scheme = "blueish"
+        sankey.plotter_args.flow_color = "hsl(230,20,70)"
+        carbonation.do_visualize = true
+
+    [cement.export]
+        path = "data/cement/output/export"
+        docs.path = "docs/cement"
+        pickle.do_export = false
```
leonieschweiger:
```suggestion
```

config/default.toml
```diff
@@ -0,0 +1,110 @@
...
+        eol_stock.do_visualize = false
+        eol_stock.over_gdp = false
+        eol_stock.per_capita = false
+        prod_clinker.do_visualize = false
+        prod_cement.do_visualize = true
+        prod_product.do_visualize = false
+        sankey.plotter_args.split_flows_by = "Stock Type"
+        sankey.plotter_args.color_scheme = "blueish"
+        sankey.plotter_args.flow_color = "hsl(230,20,70)"
+        carbonation.do_visualize = true
```
leonieschweiger:
```suggestion 
        eol_stock.do_visualize = false
        eol_stock.over_gdp = false
        eol_stock.per_capita = false
        prod_clinker.do_visualize = false
        prod_cement.do_visualize = true
        prod_product.do_visualize = false
        sankey.plotter_args.split_flows_by = "Stock Type"
        sankey.plotter_args.color_scheme = "blueish"
        sankey.plotter_args.flow_color = "hsl(230,20,70)"
        carbonation.do_visualize = false
```

config/default.toml
```diff
@@ -0,0 +1,110 @@
...
+        do_stock_extrapolation_with_time_factor = true
+
+    [plastics.visualization]
+        figures_path = "data/plastics/output/figures"
+        consumption.do_visualize = true
+        consumption.per_capita = true
+        trade.do_visualize = true
+        sankey.do_visualize = true
+        sankey.plotter_args.slice_dict.e = "C"
+        sector_splits.do_visualize = true
```
leonieschweiger:
```suggestion
        sankey.plotter_args.slice_dict.e = "C"
```

config/default.toml
```diff
@@ -0,0 +1,110 @@
...
+        do_show_figs = true
+        do_save_figs = false
+        plotting_engine = "plotly"
+        use_stock.do_visualize = true
+        use_stock.over_gdp = true
+        use_stock.per_capita = true
+        use_stock.accumulate_gdp = false
+        gdp.do_visualize = false
+        gdp.per_capita = false
+        production.do_visualize = false
```
leonieschweiger:
```suggestion
        production.do_visualize = true
```

config/default.toml
```diff
@@ -0,0 +1,110 @@
...
+        docs.do_export = true
+        docs.path = "docs/plastics"
+
+[steel]
+    [steel.model_switches]
+        do_stock_extrapolation_by_category = true
+
+    [steel.visualization]
+        figures_path = "data/steel/output/figures"
+        production.do_visualize = true
```
leonieschweiger:
```suggestion
```

config/default.toml
```diff
@@ -0,0 +1,110 @@
...
+        use_stock.do_visualize = true
+        use_stock.over_gdp = true
+        use_stock.per_capita = true
+        use_stock.accumulate_gdp = false
+        gdp.do_visualize = false
+        gdp.per_capita = false
+        production.do_visualize = false
+        trade.do_visualize = false
+        consumption.do_visualize = false
+        consumption.per_capita = false
```
leonieschweiger:
```suggestion
        consumption.do_visualize = true
        consumption.per_capita = true
```

leonieschweiger (commented):
> > We could probably declutter the default.toml a bit by agreeing on a visualization and export set for all materials that we would like to plot/export regularly in the team
> 
> I agree! @leonieschweiger could you make a suggestion?

This would be my suggestion: 
- export only pickle
- visualize stock per capita, consumption per capita, total production (I realized that plastics does not yet have a visualization for production, I will add this in a separate PR).

What do you think? Are there plots missing that you would like to regularly plot for all materials? @JakobBD @bennet21

tobiasdiez:
Apart from deciding about the default export + visualization config, this is now ready from my side.

Not as part of this PR, we can also consider to have config toggles (like `--export` or `--no-visualization`) as cli arguments. Would also be helpful for AI to have a way to run the model without opening 500 tabs ;-)

remind_mfa.py
JakobBD: That's fine, thanks!

remind_mfa.py
JakobBD: Alright!

config/default.toml
JakobBD: Just to have it written: All acceptable for me, as discussed. You decide! :-)

JakobBD:
> This would be my suggestion [...] What do you think?

Sounds good! I would just try it that way now and possibly change later if something bugs us :-)

Minor suggestion I have is to write the `slice_dict` as a dictionary. 
That way, the whole dict is always overwritten. 
I find it dangerous if e.g. the `base` part has a single region slice, and the plastics part just adds an element slice. 
The user might not intend the region slice, but miss it.  

> we can also consider to have config toggles (like --export or --no-visualization) as cli arguments. 

sounds good!

JakobBD (approved):
After merging Leonie's suggestions (and the slice_dict suggestion), seems good to merge to me! 

Thanks for that! Looking forward to the new `remind_mfa.py`, especially the all-materials option.

tobiasdiez:
Thanks everyone for your testing and numerous helpful suggestions. I've now implemented Leonie's remarks and the dict change. Hence:
image

PR #157, opened 2026-07-16
feat: Update Scenario Content, Structure, and Handling (and some other improvements)

bennet21:
(Contains all changes from #155)

# Purpose
This is a suggestion for a refactor of the scenario extrapolation implementation and represents a step towards more unified extrapolation handling in REMIND MFA.

### Scenario Inheritance
Scenarios now inherit from a `BASE.csv` that contains standard settings. Essentially all settings that were previously found in SSP2 are now found in there.

### New Scenarios
I added some new scenarios (SSP3-5 and SSP1-2 CE) with a first estimate of their top-down parameter adjustments (Saturation level, lifetime).
### Generalizing driver scenario S
With input data gen 1.6.0, floorspace became driver scenario dependent. This required a generalization of how this dimension is handled: `gdp_pop_scen` is now called `driver_scen` and every parameter that contains `S` will be sliced accoriding to the selected scenario (see `common_model`).
### Generalized make_mfa (in `common_model`)
This is needed to make a bottom-up mfa in the `cement_model`. Everything stays backwards-compatible.

### Example 
The parameter `emission_capture_rate` in plastics had a target value and a target year in the scenario csv. However, the config yaml set it to constant extrapolation, so the scenario values were not used. This makes scenario and config ambiguous - it is unclear which information is used. With the proposed extrapolation approach, the model definition only selects which parameters are extrapolated, and some modelling decisions (type, blending, split handling). The scenario decisions are now unambiguously stored in the scenario csvs.
@leonieschweiger the target row (0.5 by 2050) was removed from the scenario csv in #155, so the rate is currently held constant. If the row is re-added, the new approach will apply it: the `emission_capture_rate` then increases to 0.5, which creates some negative flows - so lmk what you intended. 

### Proposed Working Principle
The core working principle is maintained.
#### Parameter Definition
 In the model definition, you set up scenario parameters that are defined in the scenario.csvs. As before, there are
`PlainDataPointDefinition` which allows you to read in a single data point (currently used for driver scenario and saturation level)
`ExtrapolationDefinition` which allows you to read in whole arrays of data, which are converted into a flodym array and stored as a structured extrapolation instruction under `scenario_parameters[name]`. Let's walk through the settings

1. You provide a `name`
If the `name` mirrors a name of a parameter that - at the time of extrapolation - already exists in your MFA parameters, the extrapolator knows you want to use it to extrapolate this parameter. If the `name` doesn't match anything but you instead want to create a new parameter and extrapolate it directly, you can specify
2. `create_new`
This defaults to False. But it can be useful to initiate a new parameter here directly which you can use to anywhere in your MFA. This is currently used to create the `parameters["stock_factor"]` which is used to scale the stock after stock-extrapolation.
3. `type`
Here, you provide how the extrapolation should use your scenario parameter. Possible settings are either `factor` or `target`. Factor means that the extrapolation method takes your parameter and multiplies it by a value given by the scenario parameter to create an extrapolated parameter. Target means that the parameter blends from the last historic value to the scenario parameter value. 
4. `blending_function`
As before, you can specify how the transition happens. The extrapolation is now always based on a blending from the old parameter to the values of (or multiplied by) the scenario parameter. The `blending_function` determines the shape of it, defaulting to a linear transition.
5./6. `split_dimension_letter` and `split_receiver_item` may be niche applications
With the first, you can provide a dimension letter along which a split is supposed to sum up to one. Then, the extrapolation renormalizes. E.g., if you increase your share of reinforced concrete buildings, it will proportionally decrease your share of other building structures, i.e., wood, steel and masonry.
If you want the whole shift to reinforced concrete to happen through another coordinate, you can provide a receiver item. For instance, you could specify `M` (masonry) here, and all the growth in reinforced concrete buildings will be accompanied by a corresponding drop in masonry buildings.

A few more details.
- If you just want a constant extrapolation (hold last historic value constant), you can simply omit all entries except for `name`, see e.g. `ExtrapolationDefinition(name="material_shares_use_inflow")`.
- The scenario parameter name has to match one parameter at the time of extrapolation (if `create_new = False`). So you can even extrapolate parameters that are not part of the read-in.

#### Scenario csv
The scenario csv stays largely the same (apart from the structural changes in #155). You just provide `parameter`, `value` and `extra:year`, as well as indices if desired and REMIND-MFA can connect it to the settings in the definition through the parameter name. Inheritance works as before.
For constant extrapolation, you don't even have to provide information in the scenario.csv (a value without `extra:year` has no effect and triggers a warning; an `extra:year` without a `type` in the definition raises an error).
But there is one larger improvement: instead of typing a number in the `value` column you can now mix-and-match with a parameter name in the form of a string. At extrapolation, data from that parameter will be used as extrapolation target or factor. You could e.g. prepare a world-average structure split parameter in mrmfa, read it in as a normal parameter and then provide its name in the scenario csv such that the structure split is converged towards the world-average split. This can be combined with any settings in the `ExtrapolationDefinition`. The world-average structure split parameter can even itself be a parameter that is extrapolated - in that case it is automatically extrapolated before it is applied for the extrapolation of the pure structure split.

## Enabling Changes
The biggest changes are found in `parameter_extrapolation`. Here, the convoluted code has been combined into one class. This is why you don't have to provide any extrapolation class anymore. The previous base class `ParameterExtrapolation` is now the only necessary class. As before, it handles all possible dimension situations for incoming parameters: parameters that are constant in time, parameters that are only set historically, and parameters that already have a future trajectory. In its `extrapolate` method, the parameter is projected based on scenario settings: it either applies an absolute target or a relative factor. The `_renormalize_split` is applied when a `split_dimension_letter` is given. The `_description` provides some information for the assumption doc.

The `ParameterExtrapolationManager` works similarly to before, but now only passes `ExtrapolationScenarioParameter` instances on to extrapolation and orders them such that parameters used in other parameter extrapolations are extrapolated first.

Two new classes support this flow: `ExtrapolationDefinition` (in `common_definition.py`) is the structural class that lives in the model definition. `ExtrapolationScenarioParameter` (in `scenarios.py`) is the container created by `ScenarioReader` for each such definition - it carries the definition itself alongside the value array (object dtype, supporting string parameter-name endpoints alongside numbers), the `is_set` mask (distinguishing an explicit zero from an untouched entry), and the extras dict (e.g. `year`).

After extrapolation has been run, all extrapolation scenario parameter names are certain to be "real" parameters now - so you can use them via `parameters[name]`.

### Miscellaneous
- After extrapolation, parameters are rechecked with `self.check_parameters()` in the common model.
- data_blending has been refactored slightly for better external accessibility.
- The stock factor is now applied before the transition smoothing.

## Selected Bottom-up improvements
Buttom-up parameters like building structure splits are now acting on the *inflow* to the stock, instead of the stock itself. This makes much more sense as it is what societies could change, and therefore what should be reflected in scenarios. To accomplish this, while all the splits are refering to the stock, we did the following. For the historical period, we assume a constant split of the inflow, which is equal to the split of the stock. As long as the lifetimes are not dependent on the dimensions of the split, this ensures that the stock also follows this split, which is what we have data on. Going into the future, we can now gradually change these splits according to scenario assumptions. All of this requires a bit of back-and-forth calculations, which are now found in `cement_mfa_system_bottom_up.py`.


config/scenarios/base_bottom-up.csv
```diff
@@ -0,0 +1,13 @@
+parameter,models,value,index:Region,index:Stock Type,index:Function,index:Structure,description
+function_buildings_split_target,cement,0.4,IND,Res,RS,,New single-family homes make up 40% of residential floor area
```
leonieschweiger: For better readibility and in line with the previous implementation, what do you think about having only one line per parameter and use `extra:year` and something like `extra:isFactor` (True when relative, False when absolute)?

JakobBD (commented):
I like the direction of this!
However, I have qualms with all the underscores. 
After thinking about it for a while, here is what I would do: 

[caveat: I have to jump to a meeting now. Didn't have the time to spell this out very thoroughly, but let's check together]

- Create a class `ParameterExtrapolationDefinition` containing all the necessary information: 
  - name of the parameter 
  - blending function
  - target  (renamed zu value)
  - type: absolute or factor 
  - year
  - receiver (renamed to residual_item)
- Remove the blending function from the config and move it to the scenario definition instead
- optional: outsource baseline(!) parameter extrapolations to an own scenario definition file. 
- create `extra:` columns in the sceanrio definition files for all the fields above 
- Don't put the parameters you create from these definitions into the `parameters` dictionary, but into a dictionary of objects which store all the Parameters belong to one ParameterExtrapolation as attributes 
- I don't know the ParameterExtrapolation class very well, so let's think about if we can simplify my suggestion by leveraging some of the existing infrastructure there.

bennet21:
I made significant changes to the structure - see the updated PR description.

config/scenarios/SSP2.csv
```diff
@@ -1,16 +1,5 @@
...
-waste_his_exports,plastics,0,2030,,
-collection_rate,plastics,1.0,2100,,
-landfill_rate,plastics,0.0,2050,,
-mechanical_recycling_rate,plastics,0.4,2050,,
-chemical_recycling_rate,plastics,0.1,2050,,
-bio_production_rate,plastics,0.2,2050,,
-daccu_production_rate,plastics,0.01,2050,,
-emission_capture_rate,plastics,0.5,2050,,
+driver_scen,all,SSP2,,,
+stock_factor,cement,1.0,,,
```
leonieschweiger: For me it is not clear why you would not put this into base_top-down - maybe having just an SSP2 scenario as our baseline and all others inherit from there is less ambiguous than having this "base_top-down". But let's discuss this in #155

config/scenarios/base_bottom-up.csv
```diff
@@ -0,0 +1,8 @@
+parameter,models,value,extra:year,index:Region,index:Stock Type,index:Function,index:Structure,description
+function_buildings_split,cement,0.4,2060,IND,Res,RS,,New single-family homes make up 40% of residential floor area
+function_buildings_split,cement,0.4,2060,LAM,Res,RS,,New single-family homes make up 40% of residential floor area
+function_buildings_split,cement,0.4,2060,OAS,Res,RS,,New single-family homes make up 40% of residential floor area
+function_buildings_split,cement,0.4,2060,SSA,Res,RS,,New single-family homes make up 40% of residential floor area
```
JakobBD: This calls for lists in `index:` columns, which are then iterated over in the `apply` routine. 
Doesn't have to be part of this PR, though, just saying for the future

remind_mfa/common/common_definition.py
```diff
@@ -36,28 +39,80 @@ class PlainDataPointDefinition(RemindMFABaseModel):
...
+    """Dimensions for scenario data. Leave empty to extend an existing parameter without scenario data."""
+    create_new: bool = False
+    """Whether to create the extrapolated parameter instead of reading it from input data."""
+    type: Optional[Literal["factor", "target"]] = None
+    """Whether scenario values are relative factors or absolute targets."""
+    blending_function: str = "linear"
+    """Blending function to use for extrapolation. Must be one of the functions defined in `remind_mfa.common.data_blending.BLEND_TYPES`."""
+    split_dimension_letter: Optional[str] = None
+    """Only required if extrapolating a split. Ensures that along the given dimension the values sum up to 1."""
+    split_receiver_item: Optional[str] = None
```
JakobBD: To me, the word receiver is not intuitive, and I had to look up further documentation to understand what it is. "split_residual_item" seems more intuitive. But I suppose you left it intentionally, so you think otherwise?

JakobBD (approved):
[partly read, partly skimmed through source code. I haven't followed every detail of the parameter extrapolation module] 

Looks great! 
Haven't reviewed the bottom-up PR yet, but this one looks good to merge to me.

remind_mfa/common/common_definition.py
```diff
@@ -36,28 +39,80 @@ class PlainDataPointDefinition(RemindMFABaseModel):
...
+    Scenario CSV rows supply the endpoint in the `value` column, either as a number
+    or as the name of a model parameter whose values serve as the endpoint at the
+    row's coordinates (blending toward another parameter).
+    """
+
+    dim_letters: tuple[str, ...] = ()
+    """Dimensions for scenario data. Leave empty to extend an existing parameter without scenario data."""
+    create_new: bool = False
+    """Whether to create the extrapolated parameter instead of reading it from input data."""
+    type: Optional[Literal["factor", "target"]] = None
```
leonieschweiger: Would it be possible to define the type not in the definition, but in the config as an `extra:type`? I think this would make scenario settings more transparent and would simplify changing scenario configurations by externals (like REMIND users in the future).

leonieschweiger (approved):
Great work Bennet, thanks a lot! I just have one suggestion with regards to where to set the "type" of the extrapolation, let's discuss :)

remind_mfa/common/common_definition.py
bennet21: Fair, but I personally didn't think residual_item would improve understandibility. What do you think about `split_balancing_item`?

remind_mfa/common/common_definition.py
bennet21: Yes we could make it an `extra:type` and I agree that it would improve transparency/adjustability. With this change the background code would have to be adjusted to cover mixed `type` parameters as well. This is because with an in the `extra:type` you could provide different types in different coordinates. This is doable, but might make the code more complex. What do you think?

leonieschweiger: I see, I hadn't thought about this yet! But mixed `type` parameters don't necessarily need to be implemented (unless you think it is very useful) - how about throwing an error if a parameter is assigned different types in a config? We could also make this an "enhancement" labeled issue and implement it another time and just merge now as is.

bennet21: I think that's a practical compromise. I just implemented these changes. Thanks for your feedback :)

remind_mfa/cement/cement_mfa_system_bottom_up.py
```diff
@@ -0,0 +1,147 @@
...
+        self.compute_top_down_stock(td_in_use)
+        combined_stock = self.blend_stocks()
+        super().compute(combined_stock, historic_trade, stock_is_cement=False)
+
+    def compute_floorspace_stock(self):
+        """Calculate the floorspace inflow from stock change + lifetime (stock-driven)."""
+        prm = self.parameters
+        stk = self.stocks
+
+        stock = prm["floorspace"]  # TODO remove once mrmfa is fixed
```
tobiasdiez, line 41: Is there an open issue for this bug in mrmfa?

remind_mfa/cement/cement_mfa_system_bottom_up.py
```diff
@@ -0,0 +1,147 @@
...
+        7. Compute the complete MFA with the blended stock and historic trade.
+        """
+
+        self.compute_floorspace_stock()
+        self.compute_bottom_up_stock()
+        self.compute_top_down_stock(td_in_use)
+        combined_stock = self.blend_stocks()
+        super().compute(combined_stock, historic_trade, stock_is_cement=False)
+
+    def compute_floorspace_stock(self):
```
tobiasdiez, line 36: Make these other compute methods private (leading underscore)?

tobiasdiez (approved):
From a code-perspective, looks good to me. Two small remarks in the code below. Moreover, I would suggest to add a basic test for the paramater extrapolation/blending. But none of that is super important, so feel free to ignore these suggestions and just merge.

remind_mfa/cement/cement_mfa_system_bottom_up.py
bennet21: Thanks! Some of them are currently used outside the class in the cement_model (this is new), so I'll keep them public for now.

remind_mfa/cement/cement_mfa_system_bottom_up.py
bennet21: Thanks, it's [here](https://github.com/pik-piam/industry_issues/issues/24) now!

PR #158, opened 2026-07-17
refactor: iamc export

leonieschweiger:
This PR aims to streamline the iamc export of the different material models and to make adding new variables easier.
As much code as possible was moved to common_export.py, now only the mapping of flows to iamc variables is done in the material_export.py. The PR also addresses https://github.com/pik-piam/remind-mfa/issues/82
What do you think? I am happy to hear your feedback :)

TODO
- [ ] create a common .mif with the iamc output from all material models
- [ ] agree on a nomenclature - currently we are using variables from the PRISMA project, however they are quite limited and I have added new ones for plastics. Another option would be to use the CIRCOMOD nomenclature, they have more variables.

### DIFF
--- a/remind_mfa/cement/cement_export.py
+++ b/remind_mfa/cement/cement_export.py
@@ -1,120 +1,47 @@
-import pyam
-import flodym as fd
-
-from remind_mfa.common.common_export import CommonDataExporter
+from remind_mfa.common.common_export import CommonDataExporter, IamcVariable
 
 
 class CementDataExporter(CommonDataExporter):
-    def write_iamc(self, mfa: fd.MFASystem):
 
-        model = "REMIND 3.0"
-        scenario = "SSP2_NPi"
-        constants = {"model": model, "scenario": scenario}
-
-        reported_dims = ("t", "r", "s")
-
-        # cement production
-        cement_prod_by_stock_type = (
-            mfa.flows["prod_cement => market_cement"] + mfa.flows["prod_cement => sysenv"]
-        )
-        prod_df = self.to_iamc_df(cement_prod_by_stock_type)
-        prod_df["variable"] = "Production|Non-Metallic Minerals|Cement"
-        cement_prod_idf = pyam.IamDataFrame(
-            prod_df,
-            unit="t/yr",
-            **constants,
-        )
-        cement_prod_idf.aggregate(
-            variable="Production|Non-Metallic Minerals|Cement",
-            append=True,
-        )
-
-        # clinker production
-        clinker_prod_by_stock_type = (
-            mfa.flows["prod_clinker => market_clinker"] + mfa.flows["prod_clinker => sysenv"]
-        )
-        clinker_prod_df = self.to_iamc_df(clinker_prod_by_stock_type)
-        clinker_prod_df["variable"] = "Production|Non-Metallic Minerals|Cement Clinker"
-        clinker_prod_idf = pyam.IamDataFrame(
-            clinker_prod_df,
-            unit="t/yr",
-            **constants,
-        )
-        clinker_prod_idf.aggregate(
-            variable="Production|Non-Metallic Minerals|Cement Clinker",
-            append=True,
-        )
-
-        # cement demand
-        cement_demand_by_stock_type = mfa.flows["market_cement => prod_product"].sum_to(
-            reported_dims
-        )
-        demand_df = self.to_iamc_df(cement_demand_by_stock_type)
-        demand_df["variable"] = (
-            "Material Demand|Non-Metallic Minerals|Cement|" + demand_df["Stock Type"]
-        )
-        demand_df = demand_df.drop(columns=["Stock Type"])
-        demand_idf = pyam.IamDataFrame(
-            demand_df,
-            unit="t/yr",
-            **constants,
-        )
-        demand_idf.aggregate(
-            variable="Material Demand|Non-Metallic Minerals|Cement",
-            append=True,
-        )
-
-        # cement stocks
-        cement_stock_by_stock_type = (
-            mfa.stocks["in_use"].stock[{"k": "cement"}].sum_to(reported_dims)
-        )
-        stock_df = self.to_iamc_df(cement_stock_by_stock_type)
-        stock_df["variable"] = (
-            "Material Stock|Non-Metallic Minerals|Cement|" + stock_df["Stock Type"]
-        )
-        stock_df = stock_df.drop(columns=["Stock Type"])
-        stock_idf = pyam.IamDataFrame(
-            stock_df,
-            unit="t",
-            **constants,
-        )
-        stock_idf.aggregate(
-            variable="Material Stock|Non-Metallic Minerals|Cement",
-            append=True,
-        )
-
-        # cement eol
-        cement_scrap_by_stock_type = (
-            mfa.stocks["in_use"].outflow[{"k": "cement"}].sum_to(reported_dims)
-        )
-        scrap_df = self.to_iamc_df(cement_scrap_by_stock_type)
-        scrap_df["variable"] = "Scrap|Non-Metallic Minerals|Cement|" + scrap_df["Stock Type"]
-        scrap_df = scrap_df.drop(columns=["Stock Type"])
-        scrap_idf = pyam.IamDataFrame(
-            scrap_df,
-            unit="t/yr",
-            **constants,
-        )
-        scrap_idf.aggregate(
-            variable="Scrap|Non-Metallic Minerals|Cement",
-            append=True,
-        )
-
-        idf = pyam.concat(
-            [
-                cement_prod_idf,
-                clinker_prod_idf,
-                demand_idf,
-                stock_idf,
-                scrap_idf,
-            ]
-        )
-        idf.aggregate_region(
-            variable=idf.variable,
-            region="World",
-            append=True,
-        )
-        idf.convert_unit(current="t/yr", to="Mt/yr", inplace=True)
-        idf.convert_unit(current="t", to="Mt", inplace=True)
-
-        idf.to_excel(self.export_path("iamc", f"output_iamc.xlsx"))
+    def iamc_variables(self) -> list[IamcVariable]:
+        return [
+            IamcVariable(
+                variable="Production|Non-Metallic Minerals|Cement",
+                getter=lambda mfa: (
+                    mfa.flows["prod_cement => market_cement"] + mfa.flows["prod_cement => sysenv"]
+                ).sum_to(("t", "r")),
+                unit="t/yr",
+            ),
+            IamcVariable(
+                variable="Production|Non-Metallic Minerals|Cement Clinker",
+                getter=lambda mfa: (
+                    mfa.flows["prod_clinker => market_clinker"]
+                    + mfa.flows["prod_clinker => sysenv"]
+                ).sum_to(("t", "r")),
+                unit="t/yr",
+            ),
+            IamcVariable(
+                variable="Material Demand|Non-Metallic Minerals|Cement",
+                getter=lambda mfa: mfa.flows["market_cement => prod_product"].sum_to(
+                    ("t", "r", "s")
+                ),
+                unit="t/yr",
+                per="Stock Type",
+            ),
+            IamcVariable(
+                variable="Material Stock|Non-Metallic Minerals|Cement",
+                getter=lambda mfa: mfa.stocks["in_use"]
+                .stock[{"k": "cement"}]
+                .sum_to(("t", "r", "s")),
+                unit="t",
+                per="Stock Type",
+            ),
+            IamcVariable(
+                variable="Scrap|Non-Metallic Minerals|Cement",
+                getter=lambda mfa: (
+                    mfa.stocks["in_use"].outflow[{"k": "cement"}].sum_to(("t", "r", "s"))
+                ),
+                unit="t/yr",
+                per="Stock Type",
+            ),
+        ]
--- a/remind_mfa/cement/cement_export.py
+++ b/remind_mfa/cement/cement_export.py
@@ -6,14 +6,14 @@ class CementDataExporter(CommonDataExporter):
     def iamc_variables(self) -> list[IamcVariable]:
         return [
             IamcVariable(
-                variable_name="Production|Non-Metallic Minerals|Cement", # PRISMA nomenclature
+                variable_name="Production|Non-Metallic Minerals|Cement",  # PRISMA nomenclature
                 calculation_function=lambda mfa: (
                     mfa.flows["prod_cement => market_cement"] + mfa.flows["prod_cement => sysenv"]
                 ).sum_to(("t", "r")),
                 unit="t/yr",
             ),
             IamcVariable(
-                variable_name="Production|Non-Metallic Minerals|Cement Clinker", # PRISMA nomenclature
+                variable_name="Production|Non-Metallic Minerals|Cement Clinker",  # PRISMA nomenclature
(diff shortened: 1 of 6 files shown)


leonieschweiger:
I added a vibecoded script [plot_iamc_vs_validation.py](https://github.com/pik-piam/remind-mfa/pull/158/changes#diff-bf724addf578c49427a3598593cc2d6cdc49cad281a145a86fdf33e63f6079bc) that plots the iamc outputs of REMIND-MFA against the newly created validation.mif (see https://github.com/pik-piam/mrmfa/pull/65). If you prefer not to merge this, let me know.

remind_mfa/common/common_export.py
```diff
@@ -17,6 +19,22 @@
     from remind_mfa.common.common_mfa_system import CommonMFASystem
 
 
+class IamcVariable(RemindMFABaseModel):
+    """Declarative specification of a single IAMC output variable."""
+
+    variable: str
+    """IAMC variable path, e.g. "Production|Iron and Steel|Steel"."""
+    getter: Callable[..., fd.FlodymArray]
```
bennet21: As far as I understand, you want a function that takes mfa and returns the fdarray, right?
```suggestion
    getter: Callable[CommonMFASystem, fd.FlodymArray]
```

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+    def write_iamc(self, model: "CommonModel"):
+        specs = self.iamc_variables()
+        if not specs:
+            return
+        specs = self.common_iamc_variables() + specs
+
+        mfa = model.future_mfa
+        constants = {"model": self.model_name, "scenario": model.cfg.model_switches.scenario}
+
+        idfs = []
```
bennet21: For readibility, I personally prefer more descriptive names (which I don't always succeed in). So maybe this could be

```suggestion
        iamc_dataframes = []
```

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+        for variable, weight in region_weights.items():
+            idf.aggregate_region(variable=variable, region="World", weight=weight, append=True)
+
+        idf.convert_unit(current="t/yr", to="Mt/yr", inplace=True)
+        idf.convert_unit(current="t", to="Mt", inplace=True)
+        idf.convert_unit(current="t/cap/yr", to="kg/cap/yr", factor=1000, inplace=True)
+
+        idf.to_excel(self.export_path("iamc", "output_iamc.xlsx"))
+
+    def _build_idf(
```
bennet21: Same here
```suggestion
    def _build_iamc_dataframe(
```

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+        self, mfa: "CommonMFASystem", spec: IamcVariable, constants: dict
+    ) -> tuple[pyam.IamDataFrame, list[str]]:
+        """Build the IamDataFrame for a spec and return it with the variable names it produced."""
+        df = self.to_iamc_df(spec.getter(mfa))
+        if spec.per is not None:
+            df["variable"] = spec.variable + "|" + df[spec.per]
+            df = df.drop(columns=[spec.per])
+        else:
+            df["variable"] = spec.variable
```
bennet21: Would this work? [This is pure preference]
```suggestion
        df["variable"] = spec.variable
        if spec.per is not None:
            df["variable"] += "|" + df[spec.per]
            df = df.drop(columns=[spec.per])
```

remind_mfa/common/common_export.py
```diff
@@ -139,7 +255,7 @@ def export_path(self, dataset: str, filename: str = None):
 
     @staticmethod
     def to_iamc_df(array: fd.FlodymArray):
-        time_items = list(range(2025, 2101))  # TODO: more flexible
+        time_items = list(range(1950, 2101))  # TODO: more flexible
```
bennet21: Could we read this from `mfa.dims["t"].items[0]` and pass this here somehow?

remind_mfa/cement/cement_export.py
```diff
@@ -1,120 +1,47 @@
...
-            variable=idf.variable,
-            region="World",
-            append=True,
-        )
-        idf.convert_unit(current="t/yr", to="Mt/yr", inplace=True)
-        idf.convert_unit(current="t", to="Mt", inplace=True)
 
-        idf.to_excel(self.export_path("iamc", f"output_iamc.xlsx"))
+    def iamc_variables(self) -> list[IamcVariable]:
+        return [
```
bennet21, line 7: I personally prefer not to have the whole function logic in the return statement. It is understandible for me, though. :)

remind_mfa/common/common_export.py
```diff
@@ -17,6 +19,22 @@
     from remind_mfa.common.common_mfa_system import CommonMFASystem
 
 
+class IamcVariable(RemindMFABaseModel):
```
bennet21, line 23: This is a great class I think! I'd love to make it's attributes more descriptive. What do you think of:
`variable_name` instead of `variable`
`calculation_function` instead of `getter`
`split_name` instead of `per`
All of these are just suggestions, so happy for other ideas/to leave it as is.

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+            return "REMIND-MFA"
+
+    def common_iamc_variables(self) -> list[IamcVariable]:
+        """IAMC variables exported for every material.
+
+        ``Population`` is exported both as an output variable and used as the weight for
+        aggregating per-capita variables to the "World" region.
+        """
+        return [
+            IamcVariable(
```
bennet21, line 98: We don't only need pop for aggregation, bu also for reporting, right? Should we then report gdp(pc) here as well?

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+
+        Variables declared with a ``per`` dimension are summed back to their parent
+        automatically. Use this only for parents whose children are separate specs rather
+        than a ``per`` split (e.g. summing "…|Primary" and "…|Secondary" into their parent).
+        Override in subclasses.
+        """
+        return []
+
+    def write_iamc(self, model: "CommonModel"):
+        specs = self.iamc_variables()
```
bennet21: I'd make this more descriptive, too - like `iamc_vars`

scripts/plot_iamc_vs_validation.py
```diff
@@ -0,0 +1,240 @@
+"""Plot a material model's IAMC outputs against the curated validation data.
```
bennet21, line 1: I personally don't have a problem with vibe-coded plotting here. But I wouldn't review it then :)

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+            idf.aggregate_region(variable=variable, region="World", weight=weight, append=True)
+
+        idf.convert_unit(current="t/yr", to="Mt/yr", inplace=True)
+        idf.convert_unit(current="t", to="Mt", inplace=True)
+        idf.convert_unit(current="t/cap/yr", to="kg/cap/yr", factor=1000, inplace=True)
+
+        idf.to_excel(self.export_path("iamc", "output_iamc.xlsx"))
+
+    def _build_idf(
+        self, mfa: "CommonMFASystem", spec: IamcVariable, constants: dict
```
bennet21: Here you can remove the quotes for instance. Also, I'd preferer more clarity here, too :)
```suggestion
        self, mfa: CommonMFASystem, iamc_var: IamcVariable, constants: dict
```

remind_mfa/common/common_export.py
```diff
@@ -17,6 +19,22 @@
     from remind_mfa.common.common_mfa_system import CommonMFASystem
```
bennet21: These two could actually be moved out of the `TYPE_CHECKING`, I think (this was already true before your changes). So if you want, you can do that and then you can remove the quotes around them wherever they are used in this file :)

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+        """Extra parent variables to aggregate, beyond the automatic per-split parents.
+
+        Variables declared with a ``per`` dimension are summed back to their parent
+        automatically. Use this only for parents whose children are separate specs rather
+        than a ``per`` split (e.g. summing "…|Primary" and "…|Secondary" into their parent).
+        Override in subclasses.
+        """
+        return []
+
+    def write_iamc(self, model: "CommonModel"):
```
bennet21, line 129: I found this function a bit hard to understand because it contains direct handling of (important) special cases. It might be nice to move them into (a) separate private function(s). This helps potential readers to understand the gist of what's happening and refers them to specialized functions if they are more interested. Inline comments can then also become docstrings :)

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
         with open(export_path, "wb") as f:
             pickle.dump(model, f)
 
-    def write_iamc(self, mfa: "CommonMFASystem"):
-        raise NotImplementedError("Subclasses must implement write_iamc method")
+    @property
+    def model_name(self) -> str:
+        """Model identifier for the IAMC "model" column, including the REMIND-MFA version."""
+        try:
+            return f"REMIND-MFA {version('remind-mfa')}"
```
bennet21, line 87: Nice!

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,106 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+    ) -> tuple[pyam.IamDataFrame, list[str]]:
+        """Build the IamDataFrame for a spec and return it with the variable names it produced."""
+        df = self.to_iamc_df(spec.getter(mfa))
+        if spec.per is not None:
+            df["variable"] = spec.variable + "|" + df[spec.per]
+            df = df.drop(columns=[spec.per])
+        else:
+            df["variable"] = spec.variable
+        variables = list(dict.fromkeys(df["variable"]))
+        return pyam.IamDataFrame(df, unit=spec.unit, **constants), variables
```
bennet21: I just notice that the export is called IAM and not IAMC df. Should we drop the "s" everywhere, too?

bennet21 (approved):
This is such great work Leonie. Thanks a lot! This will make documenting our model results (and their progression over time as the model improves) much easier! And it is one more important step towards aligning our materials and making material specifics easier to understand!
The comments I have are basically all optional and preference based. Some of them just reflect what would've helped me a bit to understand what's going on.

remind_mfa/common/common_export.py
leonieschweiger: I think the original idea here was to define a custom export time period - should we enable this in the config? Something like:
```
export:
  iamc:
    do_export: False
    t: 1950:2100
```

scripts/plot_iamc_vs_validation.py
```diff
@@ -0,0 +1,240 @@
...
+
+For every variable that appears in both the model IAMC output and the validation ``.mif`` file,
+one PNG is produced: region panels (subplots), one colour per model, one linestyle per scenario.
```
JakobBD: what is "model" here? "source" or "material"? Maybe clarifay?

scripts/plot_iamc_vs_validation.py
JakobBD: I agree. Go wild in the scripts folder. Maybe add a note/warning in this doc string that it's un-reviewed vibe-coding. 

NB: cs2 does the same job, just with different kinds of plots and manual selection/layout for each variable.
Also added functionality, e.g. stacked bars showing multiple variables in one plot. 
I would still keep this light-weight script, though.

remind_mfa/common/common_export.py
JakobBD: NB: 
In some cases, we may want this to actually say "REMIND X.X.X". 
We could then pass an optional parameter with the version in the config or so. 
Nothing to worry about now :-)

remind_mfa/common/common_export.py
```diff
@@ -61,8 +79,148 @@ def export_model_to_pickle(self, model: "CommonModel"):
...
+                variable=variable, region="World", weight=weight, append=True
+            )
+
+    def _convert_iamc_units(self, iamc_dataframe: pyam.IamDataFrame):
+        """Convert the model's base units to the reporting units used in IAMC output."""
+        iamc_dataframe.convert_unit(current="t/yr", to="Mt/yr", inplace=True)
+        iamc_dataframe.convert_unit(current="t", to="Mt", inplace=True)
+        iamc_dataframe.convert_unit(current="t/cap/yr", to="kg/cap/yr", factor=1000, inplace=True)
+
+    def _build_iamc_df(
```
JakobBD, line 234: maybe rename either this function or `_build_iamc_dataframe`, such that the different purposes become clear from the title.

remind_mfa/common/common_export.py
JakobBD: Yes, would be great to add that in the config!

"Future years only" has the advantage that we don't have to worry about publishing proprietary historical data :-). 
We should be careful with that, maybe even throw a warning if we include historical years!

JakobBD (commented):
That's great stuff indeed! 

#### As for which variables to include: 

Sounds good to follow the Circomod template for new variables. 
This might create inconsistencies with the current prisma variables. Maybe we add a source (i.e., which template we took each variable from) to at least keep track of that. 

NB: The R package `piamInterfaces` is designed to convert REMIND mifs to project-specific templates. So if we create one comprehensive mif, we can add mappings to other variable nomenclatures over there. 

Don't do all the work of adding variables yourself now. 
Just add the ones you need, and maybe add instructions for future users how to add new variables (The circomod excel file isn't public, but we could upload it to gitlab or the cloud).

#### As for combining the three materials: 

I wouldn't worry about that now. 
It's just concatenating three files. Add a minimal script to the scripts folder, if you like.  

NB: In the long run, we may want a script that also appends these to the REMIND mif. We may want to ask RSE about that.

leonieschweiger:
> That's great stuff indeed!
> #### As for which variables to include:
> 
> Sounds good to follow the Circomod template for new variables. This might create inconsistencies with the current prisma variables. Maybe we add a source (i.e., which template we took each variable from) to at least keep track of that.
> 
> NB: The R package `piamInterfaces` is designed to convert REMIND mifs to project-specific templates. So if we create one comprehensive mif, we can add mappings to other variable nomenclatures over there.
> 
> Don't do all the work of adding variables yourself now. Just add the ones you need, and maybe add instructions for future users how to add new variables (The circomod excel file isn't public, but we could upload it to gitlab or the cloud).
> #### As for combining the three materials:
> 
> I wouldn't worry about that now. It's just concatenating three files. Add a minimal script to the scripts folder, if you like.
> 
> NB: In the long run, we may want a script that also appends these to the REMIND mif. We may want to ask RSE about that.

I added both the circomod and prisma templates to our Cloud folder under data/iamc_nomenclature. Should I add instructions for future users how to add new variables to the remind-mfa docs or just add a note to the docstring?

JakobBD:
> Should I add instructions for future users how to add new variables to the remind-mfa docs or just add a note to the docstring?

The latter!

PR #159, opened 2026-07-21
feat: add script to copy madrat input data from cluster to local directory

tobiasdiez:
helpful if you need offline access or prefer not to  overwrite the data on the cluster


README.md
```diff
@@ -44,6 +33,18 @@ Currently, all implemented models require data which is not part of the reposito
 
 The data required to run the models is planned to be made accessible in the near future.
 
+If you have access to the PIK cluster, you can obtain the data as follows:
+- Clone the data repository into the same parent directory as `remind-mfa`:
+```bash
+git clone https://gitlab.pik-potsdam.de/simson_data/remind_mfa_data.git
+```
+- Set the environment variable `MADRAT_OUTPUTFOLDER` to a folder where the madrat output data should be stored (it can be a relative path), e.g., `export MADRAT_OUTPUTFOLDER=madrat_output`. Alternatively, you can set the environment variable in a `.env` file in the main directory of the repository.
+- Then, copy the madrat output data to this folder by running
```
JakobBD: This should have a condition like "if you're not running remind-mfa on the cluster"

scripts/copy_madrat_archive.py
```diff
@@ -0,0 +1,114 @@
+import argparse
```
JakobBD, line 3: Copy the script description from the one you gave to argparse to a minimal doc string on top of the file?

JakobBD (approved):
Great, thanks! 

I have pretty low quality requirements for the scripts folder, so I'm already approving now, even though it's still a draft.

tobiasdiez:
Thanks for the review and the suggestions!

PR #160, opened 2026-07-24
feat: add export for REMIND/mrindustry

tobiasdiez:
Add an option to export data that should then be used in `mrindustry`.

Each output variable is written as a seperate csv file under `data/<material>/output/export/remind_input`.

A few ponits for consideration:
- Proposal for a better naming than "remind_input"?
- Is the naming of the variables okay?
- And most importantly, is the correct data exported?


tobiasdiez:
@leonieschweiger @bennet21 could I please get your quick first reactions, especially concerning the points mentioned above in the PR description? Thanks!

remind_mfa/steel/steel_export.py
```diff
@@ -1,8 +1,36 @@
...
 class SteelDataExporter(CommonDataExporter):
 
+    def get_remind_input_variables(self) -> list[RemindInputVariable]:
+        def steel_raw_production(mfa: fd.MFASystem) -> fd.FlodymArray:
+            """Raw steel output after production and forming losses, before trade."""
+            return mfa.flows["forming => ip_market"].sum_to(("t", "r"))
+
+        def steel_scrap(mfa: fd.MFASystem) -> fd.FlodymArray:
+            """Total scrap available after trade and before the model diverts any surplus to excess scrap."""
+            return mfa.stocks["in_use"].outflow.sum_to(("t", "r", "g"))
```
JakobBD: Should be after collection.

JakobBD:
Looks great! See my one comment. 

folder could be named `mrindustry_input`, but `remind_input` is also fine.

remind_mfa/steel/steel_export.py
tobiasdiez: Right, fixed.

bennet21:
Thanks @tobiasdiez. For cement production, I'd say you should include losses as well (prod_cement => sysenv). Is there a specific reason you call it _raw_ production? 
Another cement-related input provided to REMIND via mrindustry is clinker-to-cement ratio. That could be overwritten by the one from mrmfa and is given in `mfa.parameters["clinker_ratio"]`. Feel free to add that :)

tobiasdiez:
> For cement production, I'd say you should include losses as well (prod_cement => sysenv).

Okay, done!

> Is there a specific reason you call it _raw_ production? 

No not really, removed the `raw` prefix.

> Another cement-related input provided to REMIND via mrindustry is clinker-to-cement ratio. That could be overwritten by the one from mrmfa and is given in `mfa.parameters["clinker_ratio"]`. Feel free to add that :)

Added!

remind_mfa/plastics/plastics_export.py
```diff
@@ -1,15 +1,44 @@
...
+    def _plastic_waste(mfa: fd.MFASystem) -> fd.FlodymArray:
+        """Plastic waste available for recycling."""
+        return (mfa.flows["collected => reclmech"] + mfa.flows["collected => reclchem"]).sum_to(
+            ("t", "r", "m")
+        )
+
+    def get_remind_input_variables(self) -> list[RemindInputVariable]:
+        def plastics_production(mfa: fd.MFASystem) -> fd.FlodymArray:
+            """Plastics output after polymerization losses and before trade"""
+            return mfa.flows["polymerization => primary_market"].sum_to(("t", "r"))
```
leonieschweiger:
```suggestion
    def get_remind_input_variables(self) -> list[RemindInputVariable]:
        def hvc_input(mfa: fd.MFASystem) -> fd.FlodymArray:
            """HVC input into plastics production"""
            return mfa.flows["HVC_input => polymerization"].sum_to(("t", "r"))
```

remind_mfa/plastics/plastics_export.py
```diff
@@ -1,15 +1,44 @@
...
+
+        return [
+            RemindInputVariable(
+                name="plastics_production",
+                calculation_function=plastics_production,
+                unit="t/yr",
+            ),
```
leonieschweiger, line 34:
```suggestion
        return [
            RemindInputVariable(
                name="HVC_input",
                calculation_function=hvc_input,
                unit="t/yr",
            ),
```

leonieschweiger (approved):
Looks great, thanks! 
With regards to the exported variables: In REMIND, we only model the production of high value chemicals (HVC), the precursor of plastic production. This is why we need the HVC input into plastic production here (see my suggestions) - sorry for not specifying this earlier!

tobiasdiez:
> folder could be named `mrindustry_input`, but `remind_input` is also fine.

Thanks for the suggestion. I now went with the even simpler `mrindustry` (that it is the input for mrindustry should be clear from the context). What do you think?

tobiasdiez:
> Looks great, thanks! With regards to the exported variables: In REMIND, we only model the production of high value chemicals (HVC), the precursor of plastic production. This is why we need the HVC input into plastic production here (see my suggestions) - sorry for not specifying this earlier!

Thanks Leonie! I've now applied your suggestions.

tobiasdiez:
Thanks a lot! This is now officially ready for review.

remind_mfa/steel/steel_export.py
```diff
@@ -28,9 +56,7 @@ def iamc_variables(self) -> list[IamcVariable]:
             ),
             IamcVariable(
                 variable_name="Scrap|Iron and Steel|Steel",  # PRISMA nomenclature
-                calculation_function=lambda mfa: mfa.stocks["in_use"].outflow.sum_to(
-                    ("t", "r", "g")
-                ),
+                calculation_function=SteelDataExporter._steel_scrap,
```
JakobBD: I am afraid this change is not intended, though it would be nice. 

PRISMA's definition is pre-collection, what we need for REMIND is post-collection. 
You could call this third function `_eol` or `_scrap_pre_collection` (and the other one `_scrap_post_collection`) or so. 

Additional info: 
In the long run, I would like to have the IAMC export to include both (and be comprehensive in general), possibly with the CIRCOMOD nomenclature or our own. 
The `piamInterfaces` package can then map these onto project-specific nomenclatures.

JakobBD (approved):
Great stuff! 

Apart from the one comment, good to merge. 

NB: 
I originally would have hard-coded some functionality into this output function, where the files are directly put into the sources folder on the cluster and a new version sub-folder is created. 
But I now too prefer to move this step to a stand-alone script (as you probably intended). It is cleaner that way, and much less dangerous.

remind_mfa/cement/cement_export.py
```diff
@@ -1,15 +1,38 @@
...
+        """Cement output after production losses and before trade"""
+        return (
+            mfa.flows["prod_cement => market_cement"] + mfa.flows["prod_cement => sysenv"]
+        ).sum_to(("t", "r"))
+
+    def get_mrindustry_variables(self) -> list[RemindInputVariable]:
+        return [
+            RemindInputVariable(
+                name="cement_production",
+                calculation_function=CementDataExporter._cement_production,
```
bennet21, line 20: Why should one not use self. instead of `CementDataExporter`? :)
```suggestion
                calculation_function=self._cement_production,
```

bennet21 (approved):
Great stuff!

remind_mfa/steel/steel_export.py
tobiasdiez: I'm not sure if I follow completely. At the moment, this is only a refactoring to extract and reuse this code. It shouldn't have changed anything in the IAMC export. Do you mean that because in the future these two will be different, it doesn't make sense now to use the same function? Or do you mean that I need to export something else for mrindustry (and if so, what)?

remind_mfa/cement/cement_export.py
tobiasdiez: Here it doesn't really matter as static methods are callable in both ways.

I think it's generally a bit safer to use the class-call `CementDataExporter._cement_production` since with `self._cement_production` you might (accidentally) call also a  `_cement_production` method that is defined in a subclass of `CementDataExport`. The class-call makes it clearer that the function belongs to the class and not the instance. Here, it doesn't really matter as one probably never implement a subclass of `CementDataExporter`.

tobiasdiez:
> NB: I originally would have hard-coded some functionality into this output function, where the files are directly put into the sources folder on the cluster and a new version sub-folder is created. But I now too prefer to move this step to a stand-alone script (as you probably intended). It is cleaner that way, and much less dangerous.

When I started working on this, it was still unclear how to integrate it with mrindustry, and so I went with the most straightforward implementation. But now that Falk confirmed that a separate orchestration script is the way to go, I agree that renaming & moving it to the `source` folder on the cluster is best done there.

remind_mfa/steel/steel_export.py
JakobBD: The confusion was my bad, [this comment](https://github.com/pik-piam/remind-mfa/pull/160/changes/6f9ebabbbcdf71af60a1a360f62efec9b455c331#r3644024410) was misleading and I didn't check your fix.

IAMC export and mrindustry export need two different quantities for steel scrap. The one for IAMC export needs to remain as is. 

However, what I meant with the comment linked above is that the scrap quantity for mrindustry should be something else.

I'll just post the proposed end result here for simplicity. 

For IAMC: 
```
        def _eol_scrap_potential(mfa: fd.MFASystem) -> fd.FlodymArray:
            """End-of-life scrap available before collection and trade"""
            return mfa.stocks["in_use"].outflow.sum_to(("t", "r", "g"))
```

For mrindustry: 
```
        def _total_available_scrap(mfa: fd.MFASystem) -> fd.FlodymArray:
            """Home and new scrap, plus old scrap after collection and trade, but before the model diverts any surplus to excess scrap"""
            return mfa.flows["forming => scrap_market"] + mfa.flows["fabrication => scrap_market"] + mfa.flows["recycling => scrap_market"]
```

remind_mfa/cement/cement_export.py
```diff
@@ -1,15 +1,38 @@
...
+)
 
 
 class CementDataExporter(CommonDataExporter):
+    @staticmethod
+    def _cement_production(mfa: fd.MFASystem) -> fd.FlodymArray:
+        """Cement output after production losses and before trade"""
+        return (
+            mfa.flows["prod_cement => market_cement"] + mfa.flows["prod_cement => sysenv"]
+        ).sum_to(("t", "r"))
```
JakobBD: The documentation and function seem inconsistent. If it's after losses, why add losses? 
Which one should it be @bennet21?

JakobBD: When #164 is merged, this should be 
 
```suggestion
        """Cement output before trade and construction losses"""
        return mfa.flows["prod_cement => market_cement"].sum_to(("t", "r"))
```

tobiasdiez:
Thanks everyone for your input. I've now incorporated all open suggestions and will merge now.

PR #161, opened 2026-07-24
chore: add basic CI workflow

tobiasdiez:
Add a basic CI workflow that at the moment only verifies that the package can be installed successfully. Serves as a base for further expansion: at least adding pytest, but hopefully also actually running the mfa after copying the required input data from the cluster.


JakobBD (approved):
Great, thanks!

PR #163, opened 2026-07-27
feat: add notebook for comparing mrindustry and MFA steel data

tobiasdiez:
Adds a notebook that compares the data from MFA (as exported for mrindustry) with the current data returned by mrindustry. Currently, focused only on steel. The main functionality is to produce plots like the following (scenario + region can be selected):

Screenshot_2026-07-27_15-39-20

The purpose is two-fold: on one hand it will be helpful for me to debug issues once I start working on changing mrindustry to use the MFA data. On the other hand, it should be helpful for the discussion around the current industry energy demand (eg https://github.com/remindmodel/development_issues/issues/787). But not sure if you would like to have notebooks like this in the repo here.

Note: I've implemented it as an Marimo notebook since that's what I most familar with (and it's easier to git version), but would be happy to convert it to jupyter.


JakobBD:
Looks very helpful! 

Needs #160 for the input file to be written. 

I have no objections to Marimo. 
Within one repo, I would only have one of the two at the same time, but remind-mfa doesn't have jupyter yet, so it's fine. 
`marimo` should be added to the dependencies, I suppose? 
(The ipykernel one should actually be removed) 

I have no objections to a notebooks folder. 
The script seems related in functionality to the [plot_iamc_vs_validation script](https://github.com/pik-piam/remind-mfa/blob/main/scripts/plot_iamc_vs_validation.py). 
It may be good to merge the two somehow.

tobiasdiez:
>  `marimo` should be added to the dependencies, I suppose? (The ipykernel one should actually be removed)

Right, thanks for noticing this. Fixed.

> The script seems related in functionality to the [plot_iamc_vs_validation script](https://github.com/pik-piam/remind-mfa/blob/main/scripts/plot_iamc_vs_validation.py). It may be good to merge the two somehow.

Thanks for the pointer, I was not aware of this script. The general problem statement is similar (comparing two data sets) but the functionality is in detail then actually quite different. Perhaps one could extract from https://github.com/pik-piam/remind-mfa/blob/2a8695e898e59d7c4bd5dcc453b1acca5a7d307c/scripts/plot_iamc_vs_validation.py#L124 a method that takes two datasets as input and produces a region-based comparison plot. Personally, I would leave this idea for a follow-up PR, but can also do this here if you prefer.

JakobBD:
> Personally, I would leave this idea for a follow-up PR

Absolutely fine for me. 

Just wanted to raise awareness for future development :-) 

> the functionality is in detail then actually quite different.

True. 
But maybe in the long run, a common framework would be nice?
The plot-iamc-vs-validation seems like something where a marimo notebook would make sense, possibly with a "plot all" option.

NB: The "REMIND way" of doing this would be to create a .mif or cs4r file for all runs to be compared, and then run an adapted compareScenarios2. CS2 (comparescenarios2) has 1000 wrapping layers, so it's convenient for "plot all", but really clunky for quick customized comparisons. So I'm not advocating against an own python solution, but want to be explicit that we're building parallel structures to some extent.

JakobBD (approved):

tobiasdiez:
Thanks for the review!

> > Personally, I would leave this idea for a follow-up PR
> 
> Absolutely fine for me.
> 
> Just wanted to raise awareness for future development :-)

I've created https://github.com/pik-piam/industry_issues/issues/45 to keep track of this.

tobiasdiez:
@JakobBD here are the most important plots for the parameters coming from the mass-balance equation

Some terminology:
- lossrate = pig iron + scrap + dri / bof + eaf
- scrap part = srcap in given furnace type / all scrap
- scrap rate = scrap in given furnace type / output (bof or eaf)
- pooled regression =  nonnegative least square regression over all years + regions combined
- linear coefficients = the coefficients in the linear relation `bof = a * pig + b * srcap + c * dri` (and similar for EAF)
- neither the parameters (lossrate, scrap part) nor the linear coefficients derived from them seem to follow a clear time trend, so I just took the average of them for each region

image

image

image

Short summary:
- The approach via the mass-balance equation does capture regional difference quite well
- The coefficients/parameters do mostly make sense, except for India (negative coefficient for scrap-in-bof, coefficient greater than 1 for scrap-in-eaf). Note that the scrap usage data for India are collected from various sources, and not actually coming from BIR. 
- The mass-balance approach outperforms the regression everywhere, except for China (there the regression is actually slightly better). Especially, the performance for the EAF part is better using the mass-balance equation.

tobiasdiez:
Sorry, wrong PR.

PR #164, opened 2026-07-28
bugfix: move cement losses behind cement market

bennet21:
Cement losses refer to share of cement lost during the construction phase. This means the loss must not be applied after production, but rather after the cement market. This PR fixes that.


JakobBD (approved):
LGTM, as the pros seem to say 😉

PR #166, opened 2026-08-04
feat: add notebook for process graphs

tobiasdiez:
Adds a small notebook that generates all the process graphs for all models (for future + historic).

Needs https://github.com/pik-piam/flodym/pull/180.

### DIFF
--- a/notebooks/process_graphs.py
+++ b/notebooks/process_graphs.py
@@ -1,0 +1,33 @@
+import marimo
+
+__generated_with = "0.23.15"
+app = marimo.App()
+
+
+@app.cell
+def _():
+    import dotenv
+    import marimo as mo
+    from flodym.export import GraphvizProcessGraphPlotter
+
+    from run_remind_mfa import init_model, read_model_config
+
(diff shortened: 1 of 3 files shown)


tobiasdiez:
Todo: Currently has some issues with relative paths.

tobiasdiez:
This is working now. I had to, however, add a small feature to resolve the `data_in` path against the root directory, not the notebook location. This is https://github.com/pik-piam/remind-mfa/pull/166/commits/90268f583c3e28847047d1355f0f2192c0eee14a. Happy to move that one to a separate PR if desired.

JakobBD:
Thanks, that's great!

I don't mind which PR the `root_dir` change is part of. 
I very much like the approach, I would: 
- also use it for the export path (and other paths in the config, if there are more). 
- also use it for the `scenarios_path`, instead of making that relative to the config file? 

Then we have coherent treatment of all paths in the config file again, which improves transparency, and the scenarios_path is resolved only once after loading all config files, and not for every config file separately.

PR #167, opened 2026-08-04
chore: enable pytest in CI workflow and add pytest dependency

tobiasdiez:
With https://github.com/pik-piam/remind-mfa/pull/157, we now have tests so it's good to be able run them locally and on CI.

Fixes https://github.com/pik-piam/industry_issues/issues/30.


bennet21:
Closes https://github.com/pik-piam/industry_issues/issues/30

bennet21 (approved):
perfect, thx!

PR #168, opened 2026-08-04
feat: visualize plastics production

leonieschweiger:
- add function to visualize plastics production
- remove compare_demand plot (now replaced by validation.mif plotting script)
- add switch for visualizing material splits in consumption (previously governed by consumption switch)


JakobBD (approved):
Great, thanks!

PR #169, opened 2026-08-04
fix: move madrat_output_path check into extraction function

bennet21:
Closes https://github.com/pik-piam/industry_issues/issues/31.

`self.madrat_output_path` is now only set in the `InputCfg`. Check if the folder actually exists only happens in `extract_tar_file`.


remind_mfa/common/common_data_reader.py
```diff
@@ -49,27 +47,21 @@ def regions_filename(self) -> str:
...
-        if not check_path(env_path):
+        """Select the madrat output path (configured value, else the MADRAT_OUTPUTFOLDER env var)."""
+        path = configured_path or os.environ.get("MADRAT_OUTPUTFOLDER")
+        if not path:
             raise ValueError(
                 "No madrat output path configured. Set input.madrat_output_path or "
                 "environment variable MADRAT_OUTPUTFOLDER."
             )
-        return env_path
+        if not os.path.isdir(path):
```
tobiasdiez: Could you please move this check to extract_tar_file? In https://github.com/pik-piam/remind-mfa/blob/bad78e673d0349c8c1c09ed94280491e5cfbbda7/scripts/copy_madrat_archive.py#L49-L54 you do want to allow that the madrat folder may not exist, and then create it (and then put the madrat stuff there)

tobiasdiez (changes requested):
Thanks. I have one remark concerning the handling of the case where the folder doesn't exist. Otherwise looks good!

JakobBD (commented):
What Tobias says!
Great otherwise.

remind_mfa/common/common_data_reader.py
bennet21: Thanks! I moved the check and also made `resolve_madrat_output_path` a `@cached_property`, which avoids lazy variable setting. I have only learned about that feature now - so lmk if this is not an appropriate application.

tobiasdiez: A (cached) property belongs to the instance itself (i.e. you call it via `self.xyz`) while a static method doesn't have access to `self`. The latter is essentially a helper method that conceptually belongs to the class but otherwise could live also independently.

---

Here your change makes sense if you look only at the CommonDataReader. However, the `copy_madrat_archive` script doesn't initalizes an instance of the CommonDataReader and only calls the helper method resolve_madrat_output_path. This is now no longer working. So either you revert to a static method or move the property to the input cfg class. In my opinion, the latter would be the cleanest in terms of separation of responsibilities (the config class handles the config options, the reader how to get the data).

bennet21: You are right. I updated again - does that work for you? :)

tobiasdiez: Yes, looks good to me! Thanks!

tobiasdiez (approved):

PR #172, opened 2026-08-05
refactor: adapt Plastics MFA to new input data

leonieschweiger:
This PR does the following:

**1. Adaptation of the plastics MFA to run with the new input data (see https://github.com/pik-piam/mrmfa/pull/66).**
The historical MFA now starts with plastics production data (instead of consumption data as before), split by type (new dimension p: Plastics/Fibre/Rubber, which is essentially a broader categorization of the materials dimension m). Apparent consumption is now calculated from this production data and historical trades, resolved by type dimension p, and then split into goods and materials using the new sector_polymer_split in apparent consumption. Input data version is accordingly updated in all material models to 2.0.0.

**2. Improvement of downscaling trades to not exceed demand/supply:**
Historic exports exceed domestic production/fabrication (leading to negative consumption in the historical MFA) and historic imports exceed domestic demand (leading to negative production in the future MFA) in several cases. Previously, exports were simply capped at domestic production/fabrication and imports at demand to avoid negative flows. However, this does not allow for stopover trade (i.e. importing and re-exporting) and the different cappings in historical and future MFA led to significant deviation of historic plastics production in the future MFA from the original input data. Now, trades are scaled down only in the historical MFA such that the **net** exports do not exceed domestic production/fabrication. The respective function `cap_historical_net_exports_to_supply` is implemented in `common_mfa_system`, and the historic steel MFA now also calls it, dropping the previous implementation that also scaled down gross trades. Since the trade balancing reintroduces the scaled down excess trade, the downscaling was implemented via an iteration. 
Imports are not scaled down anymore in the future MFA in order to not exceed demand. Instead, the historical sector polymer split of the apparent consumption is adjusted already in the historical MFA to cover historical excess net final imports (function `get_historical_use_inflow_by_trade_adjusted_split`, the same as in the steel MFA, now also moved to `common_mfa_system`). To avoid historical excess net primary imports, the material split of the primary trade is adjusted in the future MFA (function `_redistribute_primary_import_excess_within_type`), in order to avoid historical negative production flows which otherwise arise because in the future MFA, primary trade and production are not only differentiated by type p (as in the historical MFA), but also material m.

**3. Improvement of stopover trade handling in the TradeExtrapolator:**
I also had to slightly refactor the stopover trade handling in the TradeExtrapolator: The previous implementation removed the stop-over volume only from `historic_second`, but not from `historic_first`, leaving historic imports (in backwards mode) inflated with stopover trade. This affects the global trade share `trd_share_glob_0` which can exceed 1 when stopover trade is included in historic impots, inflating the import share that regions with high demand growth approach to >1 in these cases, which led to scaler_second becoming negative. What the refactoring does:
- remove_stopover() subtracts the stop-over from both the imports and exports before scaling. Each flow's local trade share and the global trade share are then true shares ≤ 1, so imports cannot over-grow beyond their domestic driver and domestic supply stays non-negative.
- add_stopover() scales the stop-over separately by **global scaler (global demand in backwards mode)** (this was considered a more appropriate driver of stop-over trade than local demand) and adds it back to both future flows, keeping the re-export balanced across imports and exports.
- Adds a .maximum(0) clamp on future_first in the balance loop: a large excess could otherwise drive regional imports negative.

**4. Global sector_split_limit:**
The sector split in consumption does not differ systematically between regions according to the new input data. Therefore, the sector_split_limit is now set globally to make the stock extrapolation more transparent and easier to interpret.

**5. Minor other improvements:**
- align axes in iamc validation plotting script for per-capita plots
- plotting of total stocks per capita in the plastics visualization


JakobBD (approved):
Look great! 
Trade extrapolation refactor is definitely an improvement.
Didn't follow all your methods to the last line, but I'm sure it makes sense! 

NB: Steel has a similar problem of limiting (end-use) trade to net imports, but I think it's different enough not to be re-usable: 
In steel, the sector split is highly uncertain, so I first adapt the sector split such that end-uses with high net imports get larger shares, and only caps trade if that is not enough. 
Do you have the same situation? If so, we could merge the methods.

leonieschweiger:
> NB: Steel has a similar problem of limiting (end-use) trade to net imports, but I think it's different enough not to be re-usable: In steel, the sector split is highly uncertain, so I first adapt the sector split such that end-uses with high net imports get larger shares, and only caps trade if that is not enough. Do you have the same situation? If so, we could merge the methods.

Sector splits are definitely uncertain, as well as material splits, both in demand as well as in trade. However, this applies only to the future MFA, as in the forward calculation mode, I only have the polymer type dimension which has no such high uncertainty. In the future MFA, adapting the sector- and materials splits could be an option, however I would need to think about how to achieve this with the higher dimensionality... For the moment, this is not my priority, since I have checked that with the current implementation, the production in the future MFA only deviates marginally from the original production input parameter. Thanks for the hint though!

remind_mfa/plastics/plastics_mfa_system.py
```diff
@@ -204,6 +205,69 @@ def compute_flows(self, historic_trade: TradeSet):
...
+                net_import_agg = net_imports.sum_to(("h", "r"))
+                max_reduction = np.max((excess_agg / net_import_agg.maximum(eps)).values)
+                logging.warning(
+                    f"Historic net imports exceed domestic demand {demand.name}; "
+                    f"scaled down {len(coords)} entries:\n{detail}"
+                    f"\nNet imports reduced by up to {max_reduction:.0%} in a single region and year "
+                    f"to cap them at domestic demand."
+                )
+            # reduce imports to min(imports, demand + exports); factor in [0, 1]
+            import_factor = (imports_total - net_import_excess) / imports_total.maximum(
```
tobiasdiez: What's the purpose of the `maximum` call? If `imports_total` is zero, then this now yields effectively negative infinity as `import_factor`. Is this intended? (according to the comment you want it to be in [0,1])

remind_mfa/plastics/plastics_mfa_system.py
```diff
@@ -204,6 +205,69 @@ def compute_flows(self, historic_trade: TradeSet):
...
+            imports = historic_trade.imports
+            imports_total = historic_trade.imports.sum_over("p")
+            exports_total = historic_trade.exports.sum_over("p")
+            # net imports (imports - exports) may not exceed domestic demand
+            net_import_excess = (imports_total - exports_total - demand).maximum(
+                0
+            )  # (h, r, m[, g])
+            if not (net_import_excess.values > tolerance).any():
+                break
+            if iteration == 0:
```
tobiasdiez: This `iteration == 0` steps feels strange. For better readability I would put this part before the loop. To prevent code duplication you can extract the calculation of imports_total, net_import_exess to another local function. WDYT?

(but it's also not a big deal)

remind_mfa/plastics/plastics_mfa_system_historic.py
```diff
@@ -23,23 +27,86 @@ def compute_flows(self):
...
+        for iteration in range(50):
+            exports = trd[trade_name].exports
+            exports_total = trd[trade_name].exports.sum_to(supply.dims.letters)
+            imports_total = trd[trade_name].imports.sum_to(supply.dims.letters)
+            # net exports (exports - imports) may not exceed domestic supply
+            net_export_excess = (exports_total - imports_total - supply).maximum(0)  # (h, r, p)
+            if not (net_export_excess.values > tolerance).any():
+                break
+            if iteration == 0:
+                coords = net_export_excess.items_where(lambda x: x > tolerance)  # rows (h, r, p)
```
tobiasdiez: same remark as above

remind_mfa/plastics/plastics_mfa_system_historic.py
```diff
@@ -23,23 +27,86 @@ def compute_flows(self):
...
-        )
+            flw["fabrication => good_market"] + trd["final_his"].net_imports
+        ) * prm["sector_polymer_split"]
+
+        flw["primary_market => sysenv"][...] = trd["primary_his"].exports
+        flw["sysenv => primary_market"][...] = trd["primary_his"].imports
         flw["good_market => sysenv"][...] = trd["final_his"].exports
         flw["sysenv => good_market"][...] = trd["final_his"].imports
 
+    def scale_trade_exports_to_supply(self, trade_name: str, supply: fd.FlodymArray):
```
tobiasdiez: Is the only difference of this method here and the one above, that it uses expert vs import? If so, it may be worth trying to unify both methods.

tobiasdiez (commented):
Looks really nice. I have a few minor suggestions.

One thing I'm not sure is if those `factor = something / otherthing.maximum(epsilon)` constructions are really producing the desired effect if `otherthing` actually happens to be zero. Maybe `factor = infinity if otherthing==0 else something/otherthing` is clearer?

And sorry for the merge conflicts in the config files.

JakobBD:
> One thing I'm not sure is if those `factor = something / otherthing.maximum(epsilon)` constructions are really producing the desired effect if `otherthing` actually happens to be zero. Maybe `factor = infinity if otherthing==0 else something/otherthing` is clearer?

These constructs should usually be followed by something like 

```
baz *= factor 
```
All variables should be non-negative and `otherthing` should only be zero where `baz` is also zero.

So the correct end result for these entries of `baz` should be zero. If we set `factor` to infinity, we get errors later on. 

Of course, we could except the zero case for the whole operation including `baz *= factor`, but it seems overengineered to me. Or do you have a simple and clean solution? 

All these quantities are FlodymArrays by the way (with numpy arrays under the hood), so it would be a lot of `np.where` or so.

tobiasdiez:
If it's indeed always followed by `baz *= factor` and `baz` is zero when `otherthing` is zero, then it doesn't matter of course.

It was not clear to me that this always the case though, eg in
```python
            import_share = trd["primary"].imports / total_trade.maximum(eps)
            import_reduction = (excess * import_share).minimum(trd["primary"].imports)
```

Here it probably a bit safer to use `import_share = 0 if total_trade == 0 else trd["primary"].imports / total_trade` or the more verbose
```python
if total_trade == 0
            import_reduction = 0
else
            import_share = trd["primary"].imports / total_trade
            import_reduction = (excess * import_share).minimum(trd["primary"].imports)
```

But it might be that total_trade is never zero, and then this would be overengineering and one could just add `assert total_trade != 0`. Wasn't sure so wanted to flag it.

leonieschweiger:
> If it's indeed always followed by `baz *= factor` and `baz` is zero when `otherthing` is zero, then it doesn't matter of course.
> But it might be that total_trade is never zero, and then this would be overengineering and one could just add `assert total_trade != 0`. Wasn't sure so wanted to flag it.

Thanks for flagging this! It is definitely worthwile checking. I have checked all these constructions with `factor = something / otherthing.maximum(epsilon)` and `otherthing` is only zero when `something` is as well, so it should be fine.

remind_mfa/plastics/plastics_mfa_system_historic.py
leonieschweiger: I have discussed a bit with Jakob and will refactor these functions again in order to do all the trade capping in the historic MFA, avoiding this step in the future MFA which leads to inconsistencies between historic and future MFA for historic years. Thanks for flagging this, I will generally try to unify methods as much as possible, also between steel and plastics MFA.

remind_mfa/plastics/plastics_mfa_system.py
leonieschweiger: For me personally, the `iteration == 0` step feels quite intuitive, so if you think it's not a big deal, I would keep it that way :)

tobiasdiez: Sure, keep it!

tobiasdiez:
> > If it's indeed always followed by `baz *= factor` and `baz` is zero when `otherthing` is zero, then it doesn't matter of course.
> > But it might be that total_trade is never zero, and then this would be overengineering and one could just add `assert total_trade != 0`. Wasn't sure so wanted to flag it.
> 
> Thanks for flagging this! It is definitely worthwile checking. I have checked all these constructions with `factor = something / otherthing.maximum(epsilon)` and `otherthing` is only zero when `something` is as well, so it should be fine.

Okay, awesome!

tobiasdiez (approved):

JakobBD (approved):
Great, thanks!! 

I have a dream that one day we find a way to eliminate or improve the iteration in the net-export capping, but today is not that day.

PR #173, opened 2026-08-11
feat: extract validation.tgz

leonieschweiger:
Addresses https://github.com/pik-piam/industry_issues/issues/51.
As the preprocessing-mfa now automatically generates a .tgz with validation data for each input data generation, we can automatically extract this when running the MFA as we do for the input data.
The validation plots can then always read the most recent validation.mif from there.


JakobBD (approved):
Looks awesome and quite helpful, thanks!

PR #175, opened 2026-08-17
feat: add configuration for mrindustry export

tobiasdiez:
Adds a config file for the export for mrindustry. Only switches the region to eu21 and toggles the visualization off.

Tested via `uv run remind_mfa.py --config default --config mrindustry --model steel`.

I've considered naming the file `cluster.toml` but thought `mrindustry.toml` is a bit clearer (and future-proof).

Fixes https://github.com/pik-piam/industry_issues/issues/53.


config/mrindustry.toml
```diff
@@ -0,0 +1,6 @@
+[base]
+    [base.input]
+        region_mapping = "eu21"
+
+    [base.visualization]
+        do_visualize = false
```
JakobBD, line 6:
```suggestion
        do_visualize = false
        
    [base.export]
        do_export = true
        mrindustry.do_export = true
        pickle.do_export = false
```

Just a suggestion: 
- the first two lines are redundant to default, but I would still include them, since they should always have these values, independent of the choices in `default.toml` (which might change in the future). Also makes these requirements transparent. 
- pickle export turned off saves storage and run time.  
- I would personally set `mrindustry.do_export = false` in the default.toml. 

All optional though, your choice!

JakobBD (approved):
Great, thanks! 

See single optional comment.

config/mrindustry.toml
tobiasdiez: Thanks for the suggestion. I agree, it is worthwhile to have the `base.export` here in the `mrindustry.toml` as well.

I accepted your proposal with one modification: actually setting the pickle export to true. The compare runs script that we currently have eats pickles. So if you want to debug why the exported mrindustry data changed, it is handy to have the pickled export available as well - and is fast to generate and relatively small (okay 500mb...). Are you okay with this? 

> I would personally set mrindustry.do_export = false in the default.toml.

okay

JakobBD: Sure! I see problems with the pickle format, but I opened a [separate issue](https://github.com/pik-piam/industry_issues/issues/56) for this, so this PR is good to go!

tobiasdiez:
@JakobBD I just saw it now, but unfortunately the two submodules from https://github.com/pik-piam/remind-mfa/pull/180 actually slipped in here (in the last commit that you didn't reviewed). I'm very sorry for that.

How should we move forward with this? Revert and then decide about a proper name/place or actually integrate it properly (i.e. change the config to load from the submodule)?

JakobBD:
I've noticed, but it hasn't bothered me so far. Your choice how to move forward. 

We could still find a solution for the few remaining sources and fully integrate it in the main branch, but still leave it as a submodule in the transience branch? Or does it create problems if submodules only exist in some branches?

PR #176, opened 2026-08-17
fix: update config handling in madrat archive copy script

tobiasdiez:
I've forgot to update that script in the last big config PR. Now `copy_madrat_archive` behaves the same as `remind_mfa` when it comes to the config.

### DIFF
--- a/remind_mfa/cli/helper.py
+++ b/remind_mfa/cli/helper.py
@@ -1,0 +1,11 @@
+import typer
+
+from remind_mfa.common.config_loader import get_config_paths
+
+
+def prompt_for_config_names() -> list[str]:
+    choices = ", ".join(path.stem for path in get_config_paths())
+    entered_names = typer.prompt(
+        f"Configs (comma-separated, available: {choices})", default="default"
+    )
+    return [name.strip() for name in entered_names.split(",") if name.strip()]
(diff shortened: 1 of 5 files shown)


remind_mfa/cli/helper.py
```diff
@@ -0,0 +1,11 @@
+import typer
```
tobiasdiez, line 1: Since the `remind_mfa` folder shadows the `remind_mfa.py` file, it is not possible to reuse functions from there in the `scripts`. Thus, I've moved this little helper here under `cli`. 

I think it makes sense to also move the `remind_mfa.py` to this cli folder, so that one can simply `pip install https://github.com/pik-piam/remind-mfa` and to have the `remind_mfa` executable available. Perhaps not too relevant for devs, but might be handy on the hpc. Will do this as a follow-up if you agree.

JakobBD: Yes, that makes absolute sense!
I was having qualms about the duplicate `remind_mfa` model/package name.
And I like the option to install without cloning. 

If you do, maybe put the `run_remind_mfa` function in a different file? 
- name could be `run.py` or `remind_mfa.py` or so
- folder could be `remind_mfa` parent folder (not the repo folder, but the source folder within) or `common`. 

Optionally, [these two functions](https://github.com/pik-piam/remind-mfa/blob/ef8ff1165a7ff22482ef66418a93a9a719f15c96/remind_mfa/common/helpers.py#L16-L39) could also be moved to that file.

JakobBD (approved):
Great thanks! 

See comment regarding follow-up PR.

PR #177, opened 2026-08-19
chore: update dependencies

tobiasdiez:
Mostly updates pyam-iamc/ixmp4 by removing the fixed version constraint in `pyproject.toml`.

Strangely, the newer versions actually declare that they are incompatible with pandas 3 (and recent numpy). I've opened https://github.com/IAMconsortium/pyam/issues/997 so ask that these constraints can be removed. Not sure what that means for us: could it be that they really don't work with pandas 3, in which cas the iamc export would not work at the moment.

I also don't really know how to tests this change here, i.e. how to trigger the imac related paths.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,8 +30,8 @@ dependencies = [
     # Kaleido version: https://github.com/plotly/Kaleido/issues/134#issuecomment-1215791893
     # "kaleido==0.1.0.post1",
     "flodym>=0.8.0",
-    "pyam-iamc==3.0.0",
-    "ixmp4==0.11.1",
+    "pyam-iamc>=3.5.0",
+    "ixmp4>=0.16.4",
     "tabulate>=0.9.0",
     "dotenv>=0.9.9",
 ]
--- a/uv.lock
+++ b/uv.lock
@@ -15,16 +15,16 @@ resolution-markers = [
 
 [[package]]
 name = "alembic"
-version = "1.19.0"
+version = "1.19.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "mako" },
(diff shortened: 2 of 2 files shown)


tobiasdiez (commented):
To review: checkout & `uv sync`, then run the code that triggers imac/ixmp

JakobBD:
pandas 3 introduces substantial API breaks, so it seems plausible that a package heavily depending on pandas needs time to upgrade. 
Can't we also enforce pandas 2? IAMC export seems more important to me than pandas 3  🤭

tobiasdiez:
Right, what I find confusing is that the older versions of these packages didn't declare the pandas = 2.x constraint, so uv pulled in pandas 3. So if they were really not compatible (which is likely), then the IAMC export in `main` is currently failing (in the uv venv). 

Now, with the updated version of these iamc packages, they declare the pandas = 2.x constraint and so only pandas 2 is pulled in, see https://github.com/pik-piam/remind-mfa/pull/177/changes#diff-84321598744d84dbee2318e634c74c9aae39a1c253f1c4bd17ebf9ef2f807b11R1850-R1851.

JakobBD:
It took them two months after the pandas 3 release to add the constraint. I find that pretty quick actually, for a project of that size. 

Our problem was that they had a bug in a much older version, so we fixed the pyam version to the version before that, which couldn't know about the pandas 3 release.

It's likely that the pandas 3 changed behaviour doesn't break all pyam functionality, and that we only use unaffected parts.

But good to stay with pandas 2 for now, of course, as uv does anyways.

JakobBD (approved):

JakobBD:
Most importantly, of course: High time to update the versions, thanks for the fix!

PR #178, opened 2026-08-21
Improve robustness and stdout readability for country-level runs

JakobBD:
## Current status when attempting to run iso249:

- **Steel** now works flawlessly 
- **Plastics** runs **if** I change mass balance check behaviour in the historical MFA to "warn" instead of "error". It warns about negative production parameter and lots of negative flows, though. A lot of spikes in the plots, too, always the same regions. Probably needs further attention. 
- **Cement** fails on scenario parameters with H12 region codes. This type of problem will become quite common in the future, so I made it an [issue](https://github.com/pik-piam/industry_issues/issues/58). TBD in the JF. 

### Robust stock extrapolation 

If the minimization in the stock extrapolation algorithm fails for some region/good combination, make it fall back to the global regression.
Throw a warning which synthesizes the share of stocks that are affected by this. 

### Changed trade scaling 

When we scale down imports or exports to meet regional mass balance requirements, the global trade balancing now freezes these region/other id trades, because otherwise the global scaling would make them fail the regional mass balance instantly again. 

(E.g., if we scale down regional exports because net exports exceed local production, then global exports don't match global imports anymore. If we then scale down all imports (as we did so far), then the regional mass balance fails again. Now, we only scale down imports in all other regions instead). 

We still need iterations, but less: The mass balance in some of the other regions might now fail because we scaled down imports there. So the theoretical max number of required iterations is now n_regions, in practice much less, of course. 

I also improved consistency of the tolerances in these iterations. In the future MFA, this avoids some negative flows for me. 

### Logging updates 

- Remove info messages occurring on import of pyam package. TBD: I now changed the logger level befreo and after the pyam import. 
- keep indent for multi-line messages; I introduces a max line width to also achieve this for over-long messages where the terminal introduces line breaks. TBD if you like it. 
- remove some information in warnings, and add a reference to a debug-level message instead, to keep the info concise. 
- Add a few more status messages at major model steps

### DIFF
--- a/remind_mfa/common/common_export.py
+++ b/remind_mfa/common/common_export.py
@@ -9,7 +9,13 @@ from typing import TYPE_CHECKING, Any, Callable, Optional
 
 import flodym as fd
 import flodym.export as fde
-import pyam
+
+_root_logger = logging.getLogger()
+_prev_level = _root_logger.level
+_root_logger.setLevel(logging.WARNING)
+import pyam  # noqa: E402
+_root_logger.setLevel(_prev_level)
+del _root_logger, _prev_level
 
 from remind_mfa.common.assumptions_doc import assumptions_df, assumptions_str
 from remind_mfa.common.common_config import CommonCfg, ExportCfg
--- a/remind_mfa/common/common_export.py
+++ b/remind_mfa/common/common_export.py
@@ -14,6 +14,7 @@ _root_logger = logging.getLogger()
 _prev_level = _root_logger.level
 _root_logger.setLevel(logging.WARNING)
 import pyam  # noqa: E402
+
 _root_logger.setLevel(_prev_level)
 del _root_logger, _prev_level
 
--- a/remind_mfa/common/common_mfa_system.py
+++ b/remind_mfa/common/common_mfa_system.py
@@ -138,11 +138,13 @@ class CommonMFASystem(fd.MFASystem):
             ).values
         )
         logging.warning(
-            f"'{trade_name}': historic net exports exceed available domestic supply; "
-            f"scaled down {len(coords)} entries:\n{detail}"
-            f"\nNet exports reduced by up to {max_reduction:.0%} in a single region and year "
-            f"to cap them at domestic supply."
+            f"'{trade_name}' trade: historic net exports exceed available domestic supply; "
+            f"scaled down {len(coords)} entries in {len(by_region)} regions. "
+            f"Net exports reduced by up to {max_reduction:.0%} in a single region and year "
+            f"to cap them at domestic supply. Enable logging.DEBUG to see the affected regions "
+            "and categories."
         )
+        logging.debug(f"'Affected regions and categories:\n{detail}")
 
     def get_historical_use_inflow_by_trade_adjusted_split(
         self,
--- a/remind_mfa/common/common_mfa_system.py
+++ b/remind_mfa/common/common_mfa_system.py
@@ -66,7 +66,7 @@ class CommonMFASystem(fd.MFASystem):
         """
         trade = self.trade_set[trade_name]
         eps = sys.float_info.epsilon
-        tolerance = 100 * self._absolute_float_precision
+        tolerance = 10 * self._absolute_float_precision
         # Compare net exports and supply on the dimensions they share. The trade may be more
         # granular than the supply (e.g. per-good indirect exports vs. total fabrication supply)
         # or coarser (e.g. aggregate scrap exports vs. per-good recovered scrap), so both are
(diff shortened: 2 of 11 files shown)


remind_mfa/common/common_mfa_system.py
```diff
@@ -138,11 +140,13 @@ def _warn_historical_net_export_cap(
             ).values
         )
         logging.warning(
-            f"'{trade_name}': historic net exports exceed available domestic supply; "
-            f"scaled down {len(coords)} entries:\n{detail}"
-            f"\nNet exports reduced by up to {max_reduction:.0%} in a single region and year "
-            f"to cap them at domestic supply."
+            f"'{trade_name}' trade: historic net exports exceed available domestic supply; "
```
tobiasdiez, line 143: Nitpicking: multiline f-string (`f"""`) might be easier to read

remind_mfa/common/common_export.py
```diff
@@ -9,7 +9,14 @@
 
 import flodym as fd
 import flodym.export as fde
-import pyam
+
+_root_logger = logging.getLogger()
```
tobiasdiez, line 13: Not very important, the following might be a bit cleaner (but might not work depending on what messages you want to filter out)

```suggestion
with warnings.catch_warnings():
    warnings.simplefilter("ignore", category=...)
    import pyam
```

remind_mfa/common/trade.py
```diff
@@ -35,21 +36,41 @@ def net_imports(self):
     def net_exports(self):
         return self.exports - self.imports
 
-    def balance(self, to: str = "hmean"):
+    def balance(self, to: str = "hmean", mask_scaled: EllipsisType | np.ndarray = Ellipsis):
```
tobiasdiez, line 39: I've never worked with `EllipsisType` before, so this is more for my understanding: up in https://github.com/pik-piam/remind-mfa/pull/178/changes#diff-e1096e23fd426a09d48eea80c1082e11493b7f182cf54bc5e1b618729d4f1ba7R94-R95, you call `balance` with an array of booleans, right? Or is the argument somehow converted to the EllipsisType?

tobiasdiez (approved):
Looks great, thanks!

tobiasdiez:
CI fails as the uv lock file is now outdated (due to the added tqm dependency). If you run `uv sync`, it should update the lock file locally.

JakobBD:
> CI fails as the uv lock file is now outdated (due to the added tqm dependency). If you run uv sync, it should update the lock file locally.

I did run uv sync, but before commenting the local flodym lines again. 
I'm inclined to simply release flodym 1.0.0, I find the status quo quite annoying.

remind_mfa/common/trade.py
JakobBD: I never explicitly call it with Ellipsis, but it's the default value (which is used quite a lot). 
Hence the dual `EllipsisType | np.ndarray` type - am I doing it right?

JakobBD: And yes, I call it with `np.ndarray`, it's not converted. 

Elllipsis in numpy indexing just means "all the entries".

remind_mfa/common/common_mfa_system.py
JakobBD: Won't that introduce lots of spaces in the message to stdout?

remind_mfa/common/common_export.py
JakobBD: Doesn't work, unfortunately. It's about info level messages. 
But I now changed it to a slightly simpler version, which needs your #177, but I can wait for that :-)

remind_mfa/common/fit_stocks.py
```diff
@@ -57,13 +64,28 @@ def fit(self):
             shape=(hdims["r"].len, hdims[self.goods_dim_letter].len, self.extrapolation.n_prms)
         )
         self._n_hist = hdims["h"].len
-        for ig in range(hdims[self.goods_dim_letter].len):
-            for ir in range(hdims["r"].len):
+        ids_failed = []
+        n_r = hdims["r"].len
+        n_g = hdims[self.goods_dim_letter].len
+        iterator = np.ndindex((n_r, n_g))
+        status_bar = tqdm(
```
JakobBD, line 71: I added a status bar to this loop, which takes up more than 50 % of the overall runtime for iso249. 
If you don't like it, I can revert it again.

remind_mfa/common/common_mfa_system.py
tobiasdiez: Right, forgot the indention problem. Just leave it like it is. Sorry for the noise.

remind_mfa/common/trade.py
tobiasdiez: Thanks for the explanation, makes sense now to me.

tobiasdiez:
> I'm inclined to simply release flodym 1.0.0, I find the status quo quite annoying.

That would be nice, indeed!

